PR #13, opened 2023-04-25
Location page

### LINKED ISSUE #10: Implement location page
Screen Shot 2023-03-28 at 8 49 47 PM

kevinluo201:
### Done
* add/turn the location page into a functional component
* add leaflet to display the map interface

### Todo list
- [x] **I still don't know how to show the marker's icon correctly** but we can fix it later
- [x] The current location of the device will be fetched asynchronously. We need to wait until it's loaded and go to the location page. **Use can't go the next page**
- [x] Where should we locate the centre of the map if the user doesn't agree to fetch his/her device's location? **In Lake Ontario**

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -17,7 +17,7 @@ jobs:
       - name: Build gatbsy static frontend
         run: make build-frontend
       - name: Publish to Cloudflare Pages
-        uses: cloudflare/pages-action@v1
+        uses: cloudflare/pages-action@v1.0.0
         with:
           apiToken: ${{ secrets.CLOUDFLARE_API_TOKEN }}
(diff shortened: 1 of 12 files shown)


sonalranjit:
I think for default location we've previously used City Hall's location at `43.653220, -79.384452`

sonalranjit:
Getting an error when building the static site 
```error "window" is not available during server-side rendering. Enable "DEV_SSR" to debug this during "gatsby develop".```
Looking online for the issues led me here, seems like the leaflet component is trying to access the `window` before it exists.
https://github.com/PaulLeCam/react-leaflet/issues/45 
i'll try and look into it when i get time

sonalranjit (approved):
Awesome work @kevinluo201 thanks!

PR #17, opened 2023-06-05
Implemented Time page

sonalranjit:
implemented time page component
- added `react-datepicker` for date/time picking

### DIFF
--- a/bikespace_frontend/src/components/Time.tsx
+++ b/bikespace_frontend/src/components/Time.tsx
@@ -1,15 +1,15 @@
 import React from "react";
 import { ParkingDuration } from "../interfaces/Submission";
 import BaseButton from "./BaseButton";
-import DatePicker, { ReactDatePicker } from "react-datepicker";
-import "react-datepicker/dist/react-datepicker.css";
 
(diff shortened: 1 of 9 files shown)


bikespace_frontend/src/components/Time.tsx
```diff
@@ -1,5 +1,79 @@
...
+                duration = ParkingDuration.Overnight;
+                break;
+        }
+        props.onParkingDurationChanged(duration);
+    };
+
+    return (
+        <div id="submission-time">
+            <h2>When did this happen?</h2>
+            <input value={dateTime.toISOString().slice(0,-8)} type="datetime-local" onChange={handleDateTime}>
```
kevinluo201: I think it'll use UTC time. I also tried to do that and I found it's ridiculously difficult!

kevinluo201: I saw people use 2 libraries, moment and moment-timezone to solve this problem. It's bad, no wonder no one really like javascript 😂

kevinluo201: image

kevinluo201 (approved):
@sonalranjit I solved it by composing the date time one field by one field 😂

sonalranjit:
Hahah amazing @kevinluo201, whatever works!

PR #56, opened 2023-11-25
add basic PWA declarations

thomassth:
this should pass the minimum PWA requirement, and allow people install the web app on their devices.


sonalranjit (approved):
Cool thanks @thomassth !

PR #57, opened 2023-11-27
Sementics patch

thomassth:
- Use native `radio` and `checkbox` input
  - not completely out of react props ~~& keyboard nav not matching native~~
- Webpage titles are now populated
- Fixed a typecheck error
- Fixed button contrast, and some CSS typos


sonalranjit (changes requested):
Hey @thomassth thanks for this. Just one thing i noticed, when the issue page is rendered on iPhone 13 with safari the issue list is cutoff. Its still scrollable but it'd be great the list is rendered for these shorter screens. image

sonalranjit (approved):
Thanks @thomassth !

PR #68, opened 2023-12-16
Added summary box and issue chart to sidebar

tallcoleman:
Added summary box and issue chart to sidebar based on reference design:

image

Key notes:
- Added plotly js for the graph creation - also tried observable js but plotly js seems better-suited for the other charts we want to implement in the future and is easier to hook up interactivity (i.e. cross-filtering based on selecting parts of the chart)
- Changed the components to be class-based to ensure they all have the necessary shared methods (e.g. update method to be called by shared state)
- Shared state now has two data properties: response_data will always return the raw data from the api, whereas display_data will return data with filters applied. (Doesn't do anything right now, but will allow for interactivity later).
- Also tweaked the heading bar size slightly
- The green primary color background in the reference design for the summary box doesn't meet WCAG contrast requirements with the white text, so I had to darken it
- Accessibility, keyboard navigation, and possibly mobile interactivity for interactive plotly charts will require some work in the future. For now I've just ensured that both the bar and the bar label are click targets (since the "abandoned" issue bar is very short and hard to click).

### DIFF
--- a/bikespace_dashboard/css/map.css
+++ b/bikespace_dashboard/css/map.css
@@ -2,7 +2,7 @@
 
 #map {
     position: absolute;
-    top: 64px;
+    top: 54px;
     bottom: 0;
     left: 320px;
     right: 0;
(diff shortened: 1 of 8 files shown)


sonalranjit (approved):
Amazing! thanks @tallcoleman

PR #69, opened 2023-12-25
Added issue filtering using dashboard bar chart

tallcoleman:
Added option to filter map and graph by issue type. Filtering can be done by clicking on one of the chart bars or its corresponding y-axis label.

[Link to example gif](https://imgur.com/a/1FbaQAp)

Other changes:

- Re-organized structure to make CSS and JS files more modular
- Added a noscript tag
- Fixed a typo in map css

Known issues:
- Bar chart as input probably needs some work for accessibility & keyboard navigation, but will try to see if I can do this in a future commit.
- Bar clicks that drag slightly (common on touchpad) don't register as clicks - I think this is because plotly triggers a separate overriding event on drag. Will have to look up if there's a way to address this. Dragging clicks is not an issue for the y-axis clicks.


sonalranjit (approved):
looks great thanks!

PR #71, opened 2024-01-02
Added charts to dashboard to show reports by day of week and heatmap of parking duration vs time of day

tallcoleman:
Added two charts to the dashboard:

- Reports by day of week bar chart
- Heatmap (cross-tab) of intended parking duration vs time of day

Day of week bar chart is interactive; selecting a bar will filter all reports shown to just that day of the week. Heatmap chart is not interactive; due to complexity of this chart it may be better to offer time of day and parking duration filtering via a separate UI.

Example of how the new charts work: https://imgur.com/a/TjdHrxD

Other changes:

- Added a README for the dashboard
- "Clear Filter" button now only appears when an active filter is applied
- Changed basemap to [Thunderforest Atlas](https://www.thunderforest.com/maps/atlas/) - may not be "final" basemap but is a little cleaner than the OpenStreetMap default
- Sidebar content will now scroll on overflow
- Fixed issue where drag-clicks on charts did not register as click events
- Fixed issue with bold font rendering too bold in Safari
- Re-organized and cleaned up older code


sonalranjit (approved):
Awesome, great work @tallcoleman ! thank you.

PR #87, opened 2024-01-09
Added custom map markers and made issue colors consistent

### LINKED ISSUE #82: Set the marker colors to correspond to the colors for the issue type
Feedback from user: 
"If I can make a suggestion, not sure if possible but the colour coding for the different issues is useful – however once you get down to the individual pin, the colour dissapears. In other words once you're zoomed in to individual instances, the pins all look the same and you have to click each one to see what the issue is. Is it possible in this tool to colour code the pins themselves to match your legend?"

The default marker is the blue from leaflet. The user is asking if it'd be feasible to match the marker color to the color in the frequency chart.

Image

tallcoleman:
Fixes #82 

Added custom markers to make it easier to see issue type. Color scheme also made consistent between marker, pop-up content, and sidebar charts.

When a report has more than one issue, the priority order (hi-low) for which issue icon is rendered is as follows:

| Issue        | icon displays | co-occurrences not shown via icon                  | Rationale                                                                                                                          |
|--------------|---------------|----------------------------------------------------|------------------------------------------------------------------------------------------------------------------------------------|
| not_provided | 500 of 500    | full (54), damaged (34), other (23), abandoned (1) | Absence of parking different than issues _with_ parking; high-priority problem                                                     |
| damaged      | 336 of 370    | full (49), other (24), abandoned (6)               | High priority since likely to prevent successful parking                                                                           |
| abandoned    | 5 of 12       | other (3), full (3)                                | More likely that parking with abandoned bike is still useable compared to damaged parking                                          |
| other        | 106 of 156    | full (16)                                          | Lower priority but potentially more interesting than "full" because it may capture important use-cases not yet provided by the app |
| full         | 235 of 357    |                                                    | Lower priority since problem may only be for a limited time unlike those above                                                     |

Happy for someone to come up with better icons and colors, but thought these would be good enough for now. Also open to a discussion about changing the render order priority for issues.

I also pulled out some shared attributes for issues into a separate JS file so that things can be consistently referenced for different components.

### DIFF
--- a/bikespace_dashboard/assets/icons/icon_abandoned.svg
+++ b/bikespace_dashboard/assets/icons/icon_abandoned.svg
@@ -1,0 +1,56 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<!-- Created with Inkscape (http://www.inkscape.org/) -->
+
+<svg
+   width="130"
+   height="130"
+   viewBox="0 0 32.499999 32.5"
+   version="1.1"
+   id="svg1"
+   inkscape:version="1.2.1 (9c6d41e, 2022-07-14)"
+   sodipodi:docname="iconv4_abandoned.svg"
+   xml:space="preserve"
+   xmlns:inkscape="http://www.inkscape.org/namespaces/inkscape"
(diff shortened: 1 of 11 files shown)


sonalranjit (approved):
Awesome thanks @tallcoleman

PR #92, opened 2024-01-14
BUGFIX - fixing overlap of nav links with main app start button

tallcoleman:
Nav links were overlapping with the start submission button on iOS due to issues with how Safari on iOS calculates vh units

## Before

</img>

## After

</img>


sonalranjit (approved):
looks good, just tested on the safari with the preview link.
i had similar issues before for safari

PR #108, opened 2024-02-29
adjust submission styles for desktop UX

bennygoldman:
- set #main-content max-width to 640px
- set width of children to 90%
- other minor changes

### I did a quick QA for mobile with devtools. It might make sense to add media queries later on but I didn't want to overcomplicate it now.


sonalranjit (approved):
Looks awesome @bennygoldman ! thanks

PR #110, opened 2024-03-03
Added date filter to dashboard

### LINKED ISSUE #104: Add date filter to dashboard
Would be nice if the date filter also helped to visualize how many issues were reported in the relevant timeframes

Example: https://codepen.io/oguzhaninan/pen/gOOYxyp

tallcoleman:
Closes #104 

The new datefilter allows the user to filter the reports by date range. The filter includes useful defaults such as "last 7 days", "last 12 months", "this year" etc. and also includes a custom range option that allows the user to set any arbitrary span within the range of parking_time values returned by the API.

Demo video: [https://imgur.com/a/5yHDig6](https://imgur.com/a/5yHDig6)

## Key Changes

Three key changes have been made to implement this feature:

* Addition of luxon.js library for easier handling of timezone-aware datetimes as well as convenient date math and date range object
* Addition of new ReportFilter class in main.js which helps to standardize how components filter the report data and provides helpful utilities for managing the state of filters (especially for filters that can be modified by more than one component).
* There are now four independent filtering options for the parking_time value that can be combined as needed. (E.g. to achieve things like "reports between date x and date y only on Tuesdays")

## ReportFilter Class

Previously, the only requirement to set a filter on the SharedState class was to provide an object with a function labeled "test" that would return a true/false value that could be used to filter the displayed data based on the property matching the filter label. Some components also set other properties to help manage their own state (e.g. clearing the filter if toggled).

In implementing the date filter, there were several challenges with this system:

* Filters had to match 1:1 with one of the values of each report, meaning that it would not be possible to set more than one type of filter on the same value (this also meant that filtering on a combination of the latitude and longitude values would not be possible)
* There was no way to coordinate components that set or modified the same filters, and understanding the desired "state" of the filter would require parsing the "test" function
* The expected interface for filters was documented in the readme, but not made explicit in the code
* A significant amount of logic "close" to the API lived in each component, so if the API changes in the future, updating the dashboard code would require checking each component's code in detail

The ReportFilter class (in main.js) helps address these challenges:

* Filters now apply a test to the entire report object, allowing for the filter organization to exist separately from the list of report values. This allows for filtering based on combinations of values and multiple overlapping filters on the same underlying value.
* ReportFilters are set by specifying their desired state, and the subclass implements the needed test function based on the state value(s). This allows for consistent state management, and components can also use ReportFilter.prototype.stateEquals() to check if a desired state is the same as the filter's current state (e.g. for toggling behaviour).
* Filtering logic is now contained in the ReportFilter subclasses, instead of spread out in individual components.

## Other Changes

Other enabling changes to the dashboard:

* Organized sidebar into different panels and added navigation to switch between them.
* Clear filter button displays regardless of which sidebar panel is visible
* Added normalize and reset to CSS and also added a template sheet for non-layout styling of elements
* Organized 3rd party dependencies into "libraries" folder (people sometimes call this a "vendored" folder) and removed cdn links
* Changed "issue_attributes.js" utility file into "api_tools.js" to better reflect broader set of tools that help work with BikeSpace API data

Incidental changes:

* Changed the id of the map component from #map to #issue-map to prevent potential namespace conflict
* Added functionality to handle conditions where no data is returned by filter criteria


sonalranjit (approved):
looks amazing thank you! @tallcoleman

PR #111, opened 2024-03-03
Added parking duration filter

tallcoleman:
Thought I would add another filter to keep the date filter company :) This one filters based on parking duration.

Other changes:

* Added parking_duration_attributes to `api_tools.js`
* Added a summary box to the filter pane (to give more user feedback on how filters are affecting the number of reports returned)
* All filters now use the `ReportFilter.filterKey` attribute for consistency

Note: ignore all the old commits - I branched this from the date filter branch instead of main, so most of the older commits are from the date filter development.


sonalranjit:
When i select a date filter like `Last 30 days` and then i toggle between `Short-Term` and `Long-Term` it seems ignore the current state of the date range filter and return all submission from the beginning.
Screenshot 2024-03-04 at 2 26 44 PM

tallcoleman:
@sonalranjit filter issue fixed. Needed to specify that `filterKey` is a static method of the ReportFilter subclasses. Can you confirm it's fixed for you as well?

sonalranjit (approved):
Yupp works great! thanks @tallcoleman

PR #118, opened 2024-03-27
Added individual READMEs for each component

sonalranjit:
- Split up main README into API and frontend's own README
- Added check for python version during setup


tallcoleman (approved):
Thanks for doing this @sonalranjit I think it makes it much easier to understand the different parts. Added some edits to put in a little more contextual info.

PR #140, opened 2024-06-05
Add components label array so umami events are labelled properly

### LINKED ISSUE #135: Umami events are being mislablled by gatby's minification production
> @sonalranjit it looks like there's some JS minification that affects the events being registered in production, e.g. the Next button from the map pin page gets reported as `next-button-from-_` instead of `next-button-from-Location`

sonalranjit:
Fixed umami events not registering properly
closes #135

### DIFF
--- a/.github/workflows/test-preview.yml
+++ b/.github/workflows/test-preview.yml
@@ -50,6 +50,7 @@ jobs:
     permissions:
       contents: read
       deployments: write
+      pull-requests: write
     steps:
       - name: Checkout
         uses: actions/checkout@v3
--- a/.github/workflows/test-preview.yml
+++ b/.github/workflows/test-preview.yml
@@ -42,7 +42,20 @@ jobs:
         run: make run-pytest
       - name: Coveralls
         uses: coverallsapp/github-action@v2
-
+  test-frontend:
+    name: Lint and Test Frontend
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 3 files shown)


tallcoleman:
@sonalranjit do you think we should do it more like this so the coupling is more explicit?

```js
const orderedComponents = [
  {'component': Issue, 'label': 'issue'},
  {'component': Location, 'label': 'location'},
  {'component': Time, 'label': 'time'},
  {'component': Comments, 'label': 'comments'},
  {'component': Summary, 'label': 'summary'},
];
```

And then e.g. `orderedComponents[step].component` or `orderedComponents[step].label`

If you think it's a good idea I can push a commit

sonalranjit:
> @sonalranjit do you think we should do it more like this so the coupling is more explicit?
> 
> ```js
> const orderedComponents = [
>   {'component': Issue, 'label': 'issue'},
>   {'component': Location, 'label': 'location'},
>   {'component': Time, 'label': 'time'},
>   {'component': Comments, 'label': 'comments'},
>   {'component': Summary, 'label': 'summary'},
> ];
> ```
> 
> And then e.g. `orderedComponents[step].component` or `orderedComponents[step].label`
> 
> If you think it's a good idea I can push a commit

Yup good idea, i'll push up the suggestion

tallcoleman (approved):
Looks great, thanks @sonalranjit !

PR #144, opened 2024-06-09
Documentation updates

tallcoleman:
Added some edits and additions to the README files to make it easier to get set up with running the API.

Also addressed two issues I ran into in the past:

- Make does not work as intended if filepath has spaces - added a note about this in the root folder README (as well as some general info about how to use the Make targets)
- Swagger UI did not work in local development - issue was that the local url did not include the port number; changed it to a relative url so that it will work regardless of port number used

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,15 +15,25 @@ They are split up accordingly into their own directories:
 
 To run any of these components please see the `README` in each directory.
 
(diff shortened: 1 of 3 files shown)


sonalranjit (approved):
amazing thank youu

PR #145, opened 2024-06-12
fixed popup not opening upon refocus

hulloanson:
@xinweny Applied a workaround. I am not confident at all about its mechanism but it seems to work. Try it!


xinweny:
Works great, thanks @hulloanson !

Issue #142, opened 2024-06-08
Add landing page into react app


sonalranjit:
In inline with the effort combine all our frontend pages into one single codebase the current `bikespace_landing_page` should be migrated into `bikespace_frontend`. The current landing page is written with some static javascript libraries and html.
As part of this migration the landing page has been slightly redesigned for a more clearer navigational user experience.

See the attached figma screens:

![Image](https://github.com/bikespace/bikespace/assets/12193813/0505daf1-3abc-42e8-ad39-668655a81ed0)

![Image](https://github.com/bikespace/bikespace/assets/12193813/58c3ec5f-451b-4800-a93e-c89e07e5b000)

There are 3 call to action buttons, `Find bike parking`, `Report your parking issue` and `View the collected data`
Each button should redirect to it's corresponding endpoint.
- `Find bike parking` -> `bikespace.ca/findparking` This should redirect to a single page with a map view with all the available bikeparking data from open data.
- `Report your parking issue` -> `bikespace.ca/submissions`
- `View the collected data` -> `bikespace.ca/dashboard`

Break up the current continuous landing page:

![Image](https://github.com/bikespace/bikespace/assets/12193813/d41391d5-0134-4d08-93e1-e025308a824e)

New endpoint:
- `bikespace.ca/about` this endpoint should have the about section that is currently on the landing page.
- `bikespace.ca/privacy` this should have the content for the `How is my data used?`

tallcoleman:
@sonalranjit I added a development branch `move-frontpage-to-app`

I got a markdown page generator running using `gatsby-source-filesystem` and `gatsby-transformer-remark`. The page template is in `src/pages/{markdownRemark.frontmatter__slug}.jsx` and the test markdown file to generate a page is in `src/content/test.md`. We should be able to make other pages by adding additional markdown files.

Page template needs some work, but it's a start. Also looks like `gatsby-plugin-mdx` may also be helpful for this kind of stuff; need to look more into it.

tallcoleman:
Can also add some previews of recent blog posts using this Wordpress API endpoint: https://public-api.wordpress.com/rest/v1/sites/147310791/posts

PR #146, opened 2024-06-12
Epic: Merge bikespace_landing_page and bikespace_dashboard into bikespace_frontend

### LINKED ISSUE #133: Add dashboard layout into app
Part of the merging dashboard into react app effort

Move what we have here:

image

into the `bikespace_frontend` project

hulloanson:
# Reasons

- `bikespace_dashboard`'s state management was getting out of hand and was recreating a lot of what `react` and ecosystem has to offer
- `bikespace_landing_page` has a new design to implement
- easier styles and components reuse
- opens more possibilities to other user flows e.g. report issue in the "dashboard" map

# Goals

- [ ] Implement Bai's landing page design in `bikespace_frontend` (#142)
- [x] Re-implement `bikespace_dashboard` in `bikespace_frontend` (#134 and ongoing changes in `develop`)
- [ ] Configure deployment and gatsby settings so existing entrypoints (app.bikespace.ca, dashboard.bikespace.ca and bikespace.ca) are still usable

`develop` serves as the base of all these changes and when the time is right, we can merge (part of) the changes to `main` and deploy them.


hulloanson:
Tagging @sonalranjit @tallcoleman @xinweny @AidanNPH for discussion

tallcoleman:
@hulloanson @xinweny should I merge the [move-frontpage-to-app](https://github.com/bikespace/bikespace/tree/move-frontpage-to-app) branch into develop? Idea was to use some of gatsby's markdown plugins so that the page content could be written in markdown files.

tallcoleman:
Also let me know if I should add in the API key for the Thunderforest base map that [dashboard.bikespace.ca](https://dashboard.bikespace.ca/) currently uses.

hulloanson:
> @hulloanson @xinweny should I merge the [move-frontpage-to-app](https://github.com/bikespace/bikespace/tree/move-frontpage-to-app) branch into develop? Idea was to use some of gatsby's markdown plugins so that the page content could be written in markdown files.

We should! @xinweny has made quite some changes with `bikespace_frontend`

hulloanson:
> Also let me know if I should add in the API key for the Thunderforest base map that [dashboard.bikespace.ca](https://dashboard.bikespace.ca/) currently uses.

Is the API key the one used here?

https://github.com/bikespace/bikespace/blob/develop/bikespace_frontend/src/components/dashboard/map/Map.tsx#L31

xinweny:
> Also let me know if I should add in the API key for the Thunderforest base map that [dashboard.bikespace.ca](https://dashboard.bikespace.ca/) currently uses.

That would be great! I was wondering why the map styles looked different when I was integrating into React, I'm guessing this could likely be why.

tallcoleman:
> > Also let me know if I should add in the API key for the Thunderforest base map that [dashboard.bikespace.ca](https://dashboard.bikespace.ca/) currently uses.
> 
> Is the API key the one used here?
> 
> https://github.com/bikespace/bikespace/blob/develop/bikespace_frontend/src/components/dashboard/map/Map.tsx#L31

Turns out it was in there already, the two tile layers were just overlapping. Committed a change so it picks just one.

tallcoleman:
## Items to polish up before merging react dashboard into main

Mobile view tested via cloudflare preview link on iOS; desktop tested via Firefox on MacOS

### High Priority
- [x] On mobile - map credits z index is higher than menu drawer
- [x] On mobile - CSS is incorrectly calculating viewport height - causes drawer to appear to disappear when closed
- [x] Date filter selection and description field do not update with clear filter button
- [x] On mobile - Styling for panel nav, filter components, text in feed - should not be blue
- [x] On mobile - Opening drawer on feed view covers entire page instead of half
- [x] Date filter applied for last 7 appears to be incorrect - first date in described range is excluded?

### Med Priority
- [x] Styling for leaflet clusters - is currently set at default; previous dashboard had custom styling to align with brand colours
- [x] Problem type frequency graph should retain order when filter applied via click
- [x] Clear filter button does not reset styling on problem type frequency
- [x] Frequency by day graph should update style instead of zeroing out values when click to filter is applied
- [ ] URL specifiers for certain views or specific reports (previously implemented with hash(tag) router) not included – should at the very least include for specific reports
- [x] Styling logic for map pins and issue "chip" colours appears to be inconsistent and/or incorrect in both the map and the feed 

### Low Priority
- [x] Clear filters button may need additional vertical padding
- [x] Add a loading state for components so that they don't show "No Data" during load (not super noticeable if API response/internet is fast)

tallcoleman:
Added one more issue to the fix list - "Styling logic for map pins and issue "chip" colours appears to be inconsistent and/or incorrect in both the map and the feed"

Examples:

Screenshot 2024-07-23 at 9 56 19 PM

Screenshot 2024-07-23 at 9 59 42 PM

Correct examples from existing dashboard:

Screenshot 2024-07-23 at 10 00 37 PM

Screenshot 2024-07-23 at 10 00 51 PM

xinweny:
> Added one more issue to the fix list - "Styling logic for map pins and issue "chip" colours appears to be inconsistent and/or incorrect in both the map and the feed"
> 
> Examples:
> 
> Screenshot 2024-07-23 at 9 56 19 PM Screenshot 2024-07-23 at 9 59 42 PM
> Correct examples from existing dashboard:
> 
> Screenshot 2024-07-23 at 10 00 37 PM Screenshot 2024-07-23 at 10 00 51 PM

Fixed!

xinweny:
I just introduced the following packages to handle the dashboard app state:
1. [Tanstack Query](https://tanstack.com/query/latest/docs/framework/react/overview) - tracks asynchronous state (fetching data from API to client, fetching all the submissions in our case)
2. [zustand](https://github.com/pmndrs/zustand) - tracks local state (e.g. submission filters, tabs, focused submission ID, etc.)

Tanstack Query makes it easier to introduce loading components like skeletons to specific components via the `isLoading` attribute from the return value of `useQuery()`.

Additionally, the code reads a bit cleaner without all the wrappers from using React Context with zustand.

tallcoleman:
Added a fix for the iOS 100vh issue, but did notice one more problem -- clicking on an issue in the feed that is part of a map cluster does not open the cluster

Reference for fix to iOS 100vh issue: https://stackoverflow.com/questions/37112218/css3-100vh-not-constant-in-mobile-browser

xinweny:
> Added a fix for the iOS 100vh issue, but did notice one more problem -- clicking on an issue in the feed that is part of a map cluster does not open the cluster
> 
> Reference for fix to iOS 100vh issue: https://stackoverflow.com/questions/37112218/css3-100vh-not-constant-in-mobile-browser

Thanks for the fix @tallcoleman!

I just pushed a fix to increase the zoom level so the map clusters now open up when the point is focused on.

tallcoleman:
I think we might need to find another way to get the clusters to open; when I tried it, it opens the clusters, but it seems to be too zoomed in too far for leaflet to show the tiles properly.

Screenshot 2024-08-03 at 10 21 24 AM

> Thanks for the fix @tallcoleman!
> 
> I just pushed a fix to increase the zoom level so the map clusters now open up when the point is focused on.

xinweny:
> I think we might need to find another way to get the clusters to open; when I tried it, it opens the clusters, but it seems to be too zoomed in too far for leaflet to show the tiles properly.
> 
> Screenshot 2024-08-03 at 10 21 24 AM
> > Thanks for the fix @tallcoleman!
> > I just pushed a fix to increase the zoom level so the map clusters now open up when the point is focused on.

Just pushed another fix, adjusted some parameters like `maxZoom` on the `<MapContainer />` component and added the `focusedId` to the `useEffect()` calling `map.invalidateSize()`, so this is called when the focused submission id changes. The tiles do load now, although quite slowly for me on Chrome - if anyone figures out a better fix that would be more than welcome.

tallcoleman:
Spent some time working on this to-do this weekend:

> URL specifiers for certain views or specific reports (previously implemented with hash(tag) router) not included – should at the very least include for specific reports

The primary use case for this is to link to a specific issue, e.g. like this in the existing dashboard: https://dashboard.bikespace.ca/#feed?view_all=1&submission_id=1387 [1]. We used these links in the report of damaged bike parking locations we sent to the City in April.

I played around with implementing URL parameter navigation and had some success, but I think the proper way to implement it without making things buggy is to:

* Switch `useQuery` to `useSuspenseQuery` ([docs here](https://tanstack.com/query/latest/docs/framework/react/guides/suspense))
* Wrap query-dependent components in a [Suspense component](https://react.dev/reference/react/Suspense)
* Add loading fallbacks for those query-dependent components (e.g. replacing the current `queryResult.isPending` logic in `DashboardPage` and `ReportSummary`)
* `Map` components might still have some bugs to iron out because Leaflet has its own events system, but we'll see...

Thoughts?

[1]: Although the old dashboard uses the URL hash parameters for state navigation, it seems like support for regular parameters (e.g. `/dashboard?tab=filters` or `/dashboard?id=1387`) is better in next, so I went with regular parameters instead.

tallcoleman:
Small improvement suggested by a user we could implement:

- [ ] Match issue type description to input instead of backend value

User feedback: "It's weird/confusing that when I selected the issue "parking not provided", on the confirmation page it said "none provided" as the issue. I'd expect the wording on the confirmation page to be similar to what I selected."

Example:
Screenshot 2024-08-17 at 11 57 26 AM
Screenshot 2024-08-17 at 11 57 47 AM

sonalranjit:
@tallcoleman should we get this in? ~i can perform the DNS changes on the cloudflare once this is deployed to forward dashboard.bikespace to point to app.bikespace.ca/dashboard~ actually might not be to possible to do this on the DNS end. I think we'll have to update the landing page code and the frontend to use the new url for dashboard app.bikespace.ca/dashboard/

sonalranjit (approved):
LGTM

PR #154, opened 2024-06-15
Merge content page template test into develop

tallcoleman:
I got a markdown page generator running using `gatsby-source-filesystem` and `gatsby-transformer-remark`. The page template is in `src/pages/{markdownRemark.frontmatter__slug}.jsx` and the test markdown file to generate a page is in `src/content/test.md`. We should be able to make other pages by adding additional markdown files.

Page template needs some work, but it's a start. Also looks like `gatsby-plugin-mdx` may also be helpful for this kind of stuff; need to look more into it.


hulloanson:
works well!

Changed the generated page to index and link them right to the submission and dashboard page.

xinweny:
I'm unable to run `make run-frontend` after pulling the new changes, got the following error:

```
Error: No native build was found for platform=darwin arch=arm64 runtime=node abi=120 uv=1 armv=8 libc=glibc node=21.6.2
    loaded from: /Users/xinweny/code/bikespace/bikespace_frontend/node_modules/lmdb and package: @lmdb/lmdb-darwin-arm64
```

I tried removing and reinstalling `node_modules` but the same error keeps coming up, I'm wondering if it might have to do with the Node version on my computer?

xinweny:
Based on the solutions [here](https://stackoverflow.com/questions/73290680/error-no-native-build-was-found-for-platform-win32-arch-x64-runtime-node-abi-10), I had to locally set `bikespace_frontend`'s Node version, which I set to v18 by `npm install node@18 --save-exact`, which fixed the error above.

The next error I encountered was:
```
Error: No prebuild or local build of @parcel/watcher found. Tried @parcel/watc
  her-darwin-arm64. Please ensure it is installed (don't use --no-optional when 
  installing with npm). Otherwise it is possible we don't support your platform 
  yet. If this is the case, please report an issue to https://github.com/parcel-
  bundler/watcher.
```

I found a fix for this [here](https://github.com/parcel-bundler/watcher/issues/152), which required me to do
```
npm install -O @parcel/watcher-darwin-arm64
npm dedupe
```

Now `make run-frontend` can build successfully. However I'm not sure if this is an ideal solution as it requires downgrading Node, was wondering if there might be another better fix for this?

PR #156, opened 2024-06-17
Fix frontend build issues with ARM64 architecture and Node v21

xinweny:
I should have probably opened a separate PR here for this, my bad!

To summarize again, I was having trouble running `make run-frontend` on my M2 Mac  after pulling the latest changes on `develop`. I encountered the following errors:

1. 
```
Error: No native build was found for platform=darwin arch=arm64 runtime=node abi=120 uv=1 armv=8 libc=glibc node=21.6.2
    loaded from: /Users/xinweny/code/bikespace/bikespace_frontend/node_modules/lmdb and package: @lmdb/lmdb-darwin-arm64
```
Using the solution found [here](https://stackoverflow.com/questions/73290680/error-no-native-build-was-found-for-platform-win32-arch-x64-runtime-node-abi-10), I fixed this by downgrading my Node version from v21 to v18, by `npm install node@18 --save-exact`.

2.
```
Error: No prebuild or local build of @parcel/watcher found. Tried @parcel/watc
  her-darwin-arm64. Please ensure it is installed (don't use --no-optional when 
  installing with npm). Otherwise it is possible we don't support your platform 
  yet. If this is the case, please report an issue to https://github.com/parcel-
  bundler/watcher.
```
Using the solution [here](https://github.com/parcel-bundler/watcher/issues/152), I ran the following:
```
npm install -O @parcel/watcher-darwin-arm64
npm dedupe
```

Which allowed `make run-frontend` to run successfully. However this would require manual downgrading of Node and installing optional dependencies, so I'm not sure if there might be better ways around this.

### DIFF
--- a/bikespace_frontend/package-lock.json
+++ b/bikespace_frontend/package-lock.json
@@ -32,6 +32,7 @@
         "leaflet.locatecontrol": "^0.81.1",
         "leaflet.markercluster": "^1.5.3",
         "luxon": "^3.4.4",
+        "node": "18.20.3",
         "plotly.js-dist-min": "^2.33.0",
         "punycode": "^2.3.1",
         "react": "^18.2.0",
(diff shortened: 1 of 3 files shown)


xinweny:
Thanks @tallcoleman for the following suggestion - running `make clean` clears these errors without needing the solution above.

Closing for now.

PR #171, opened 2024-08-14
Move tab and focused submission ID to URL param state

xinweny:
Move tab and focused submission ID out of the global store state and into the URL param state for ease of sharing different submissions to be focused on. Tab and ID state are manipulated only through URL params rather than going through the global store first, to increase simplicity.

Also increase consistency of pan -> zoom to a certain submission even through clusters by adjusting the animation duration, open popup delay time and zoom factor - but there is probably a better solution.

The tab state can eventually be moved back into the global store easily if needed.


tallcoleman (approved):
Thanks for working on this Xin Wen! Go ahead and merge; if there are any issues to address from these changes I'll note them in the main `develop` pr.

PR #173, opened 2024-08-23
Merge in City damage report 202408

tallcoleman:
Script changes:

- Converted to standalone script instead of jupyter notebook and re-organized file structure
- Added new functionality to the script to only show damage reports within a certain date range. See the new README for details.

Report features:

- Includes all relevant damage reports from 2023 onwards (20 in total, 13 of them are net new compared to the previous report). All of the 20 appear to be related to post and rings, no racks.
- Report includes photos and confirmed asset IDs (to link to City open data) from surveys
 
Report changes:

- The MatchesDisplay tab now includes survey photos as well as the thumbnail map indicating the location. Photos are pulled from the `/photos` directory based on name and resized before insertion.
- New field “City_Problem_Type” indicates the main issue reported, e.g. “Loose/unsecured”, “Post with no ring”
- New fields “Check_311_Date” and “Check_311_Notes” to record whether there are 311 reports for the issue already (though the utility of this is limited, since the City 311 page only shows 311 reports from the last 30 days at most)
- New MatchesTable tab shows the report with the first (nearest or confirmed) linked data point from the City


.gitignore
sonalranjit, line 1: you probably want to ignore this change until the `develop` branch is merged into `main` to avoid conflicts

sonalranjit (changes requested):
Thanks @tallcoleman , awesome to have everything in one place. I've just mentioned the changes for the `.gitignore` other than that we should be good to merge

.gitignore
tallcoleman: Fixed by #146

tallcoleman:
Did some additional clean up of this script:

- Added [uv](https://docs.astral.sh/uv/) as an python/dependencies manager
- Separated out tests folder
- Updated date parsing to match current api format (see: #184 )
- Updated dashboard link generation to point to new dashboard

sonalranjit (approved):
Coool! thanks @tallcoleman

PR #179, opened 2024-09-14
Added per-issue columns to csv request type

tallcoleman:
In addition to text list of which issues were selected for a submission, adds one column per issue indicating whether that issue was selected with true/false.

Have done this as a small PR in case you feel this is out of scope or adds too much complexity


tallcoleman:
Output looks like this:

Screenshot 2024-09-14 at 12 30 42 PM

sonalranjit (approved):
thank you @tallcoleman

PR #183, opened 2024-09-19
Migrate map from leaflet to maplibre and supercluster

xinweny:
Hi all,

I've been playing around with `react-map-gl` and `maplibre-gl` to replace `react-leaflet` as discussed in previous Hacknight meetings, and using `supercluster` for clustering. So far, the API interface is smoother and easier to coordinate map events, and with fewer map library dependencies. Some major changes:

1. 2 types of map markers for the dashboard - normal point markers and cluster markers. The parameters for clustering still need to be tweaked (e.g. cluster marker styling, radius thresholds, cluster marker base size, etc.).
2. Marker behaviour when URL param `submission_id` is changed is set within the `MapMarkers.tsx` components, via several `useEffect()`s. If the submission is within a cluster, the cluster will automatically be zoomed in to the minimum zoom level required to be opened. Maybe there is a better way to do this directly via the `MapGL` event handler props (e.g. `onMoveEnd`, `onZoomEnd`) but I haven't messed around with it too much aside from using it to set initial view state.
3. Removal of `MapHandler.tsx` "components", as all event handling/useEffect behaviours are handled within the marker/map/popup components themselves.
4. Geolocation functions are using `react-map-gl`'s `GeolocateControl`.
5. Migration to `react-map-gl` is also done for the submission form's `Location.tsx` component - this was much simpler to do

Also the behaviour when programmatically moving/zooming the map still needs to be fine-tuned to match what we had before (especially for mobile view) - would love to hear what you guys think. :)


sonalranjit:
Whoa very cool @xinweny the location screen on the reporting app functions very smoothly.
I think the only thing i can point out is the different in the map icons with the cluster and un clustered points. It'd be easier on the eyes i think if the min size of the clustered points is the same the unclustered so that the unclusttered points don't seem soo much bigger than the clusters.
Screenshot 2024-09-19 at 2 12 45 PM

sonalranjit (approved):

PR #184, opened 2024-09-21
Date fix for current dashboard

tallcoleman:
API date format has changed with one of the recent commits:

Old format example: `Sun, 13 Aug 2023 12:36:59 GMT`
New format example: `2024-09-20 1:00:00` (newer entries) or `2024-09-20 03:28:24.632000` (older entries)

Have changed parsing code in the current dashboard to ensure dates are still parsed correctly.

`develop` branch also needs fixing (dates are getting parsed, but in the wrong timezone, e.g. issue id 1407 should display as 8:45am instead of 12:45am).

We should also consider adding end-to-end testing to our test suite to catch integration issues between the API and frontend.

I also updated the linter settings to a more recent ECMA version.

### DIFF
--- a/bikespace_dashboard/.eslintrc.js
+++ b/bikespace_dashboard/.eslintrc.js
@@ -13,7 +13,7 @@ module.exports = {
   plugins: baseConfig.plugins.filter(notNode),
   parserOptions: {
     sourceType: 'module',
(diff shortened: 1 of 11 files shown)


sonalranjit (approved):
thanks @tallcoleman

PR #189, opened 2024-10-06
Dashboard UI cleanup

tallcoleman:
Made a couple tweaks to fix some edge cases with the new dashboard UI:

When loading a url with the issue id already specified in the parameters:

- prevents geolocate from moving the map away from the issue pin
- switches to the 'feed' tab on load when below the mobile breakpoint

Mobile breakpoint:

- Standardized to 1024px
- Changed hard-coded instances in scss to use central variable
- Single use of breakpoint in TSX component also inherits from the central scss variable


xinweny:
Looks clean, thanks @tallcoleman!

xinweny (approved):

PR #191, opened 2024-10-17
Update landing page css to match figma designs

sonalranjit:
- Updated new landing page design to match figma design
- Updated node dependency versions
Screenshot 2024-10-17 at 6 20 19 PM

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -33,7 +33,7 @@ endif
 
 clean:
 	rm -rf $(VENV)
-	cd $(BIKESPACE_FRONTEND_DIR) && npm run clean && rm -rf node_modules
+	cd $(BIKESPACE_FRONTEND_DIR) && rm -rf node_modules && rm -rf .next && rm -rf build && rm -rf out && rm -rf coverage
 
 $(VENV): check-python-version $(BIKESPACE_API_DIR)/requirements.txt
 	python3 -m venv $(VENV)
--- a/bikespace_frontend/src/components/landing/content-blocks/content-blocks.module.scss
+++ b/bikespace_frontend/src/components/landing/content-blocks/content-blocks.module.scss
@@ -18,7 +18,6 @@
   margin-bottom: 2rem;
   max-width: $wrapper-full-width;
   flex: 2 1 400px;
-  padding: 5%;
 
   .heroTagline {
     font-size: 2rem;
@@ -59,7 +58,6 @@ a.buttonLink {
   &:hover {
     border-color: $color-primary;
   }
-  padding: 1.5rem;
   width: 125%;
   gap: 1rem;
 }
--- a/bikespace_frontend/src/components/landing/content-blocks/content-blocks.module.scss
+++ b/bikespace_frontend/src/components/landing/content-blocks/content-blocks.module.scss
@@ -17,11 +17,13 @@
   margin-top: 2rem;
   margin-bottom: 2rem;
   max-width: $wrapper-full-width;
+  flex: 2 1 400px;
 
   .heroTagline {
-    font-size: 2rem;
+    font-size: 2.5rem;
     font-weight: bold;
     text-wrap: wrap;
+    gap: 1rem;
   }
 
   .heroImage {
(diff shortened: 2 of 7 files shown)


Makefile
tallcoleman, line 1: Great update!

bikespace_frontend/src/components/landing/content-blocks/content-blocks.module.scss
```diff
@@ -56,15 +59,28 @@ a.buttonLink {
   &:hover {
     border-color: $color-primary;
   }
+  padding: 1.5rem;
```
tallcoleman: This rule on line 62 doesn't need to be added again since it's already specified above; for lines 63-64, I would move above the `&` specifiers. I think there's another one like this too – was generally trying to follow the pattern of main styles first and then conditional styles added with the `&` specifiers after.

bikespace_frontend/src/components/landing/content-blocks/content-blocks.module.scss
```diff
@@ -17,11 +17,14 @@
   margin-top: 2rem;
   margin-bottom: 2rem;
   max-width: $wrapper-full-width;
+  flex: 2 1 400px;
+  padding: 5%;
```
tallcoleman: If you want more padding between the content and screen edge, I would suggest adjusting `bikespace_frontend/src/components/landing/landing-layout/landing-layout.module.scss` (line 11).

There's also a knock-on effect that with the greater padding for this component, it causes the hero button text to break into two lines on mobile:

bikespace_frontend/src/components/landing/content-blocks/content-blocks.tsx
tallcoleman, line 1: For this one I would try to avoid using `<br />` elements for spacing. The changes to the flex rules have also made it not work quite right on mobile — the image gets very small and stays in 2-col format instead of responsively dropping below the buttons.

bikespace_frontend/src/components/landing/footer/footer.module.scss
tallcoleman, line 1: This is a great improvement, especially on mobile!

bikespace_frontend/src/components/landing/landing-layout/landing-layout.module.scss
```diff
@@ -10,6 +10,8 @@ $even-lighter-grey: #EEE;
   margin: 0 auto;
   padding: 1rem;
   max-width: $wrapper-max-width;
+  display: flex;
```
tallcoleman: I would try and avoid doing this and control the justification of content on a block by block basis instead of at the page level.

tallcoleman (changes requested):
Hi Sonal, thanks for taking a look at the design – have left some changes to review, most important is that the layout still work properly on mobile.

If you came across anything confusing when you were trying to figure out the frontend folder organization, feel free to add it to the README in the `/bikespace_frontend` folder.

xinweny:
Thanks and great work @sonalranjit! I think @tallcoleman has covered most of the important points.

bikespace_frontend/src/components/landing/content-blocks/content-blocks.module.scss
sonalranjit: Updated

bikespace_frontend/src/components/landing/landing-layout/landing-layout.module.scss
sonalranjit: Updated

bikespace_frontend/src/components/landing/content-blocks/content-blocks.module.scss
tallcoleman, line 1: @sonalranjit on this one I added one small tweak - removed the `width: 125%;` rule since it was causing the buttons to overflow the parent container on mobile. Before and after:

Screenshot 2024-10-26 at 11 01 13 AM

Screenshot 2024-10-26 at 11 01 48 AM

tallcoleman (approved):
Thanks @sonalranjit !

PR #192, opened 2024-10-18
React hook form

xinweny:
I did some refactoring on the submission form as the state management of the form variables via `useState` was getting out of hand, especially as all the form fields/form state were all lumped together and too many get/set state props being passed around everywhere.

Mainly, I added the `react-hook-form` package, using `zod` for validation, which I have used in quite a few of my previous projects - I have found them very useful for creating React forms, especially complex ones. This allows separation of concerns between handling form fields, form state changes, client-side validation and error handling.

More information on React Hook Form documentation [here](https://www.react-hook-form.com/), and for [zod](https://zod.dev/) here.

I will need to update the test files as well to address this change.

If you guys like the feel of the packages, we could also consider refactoring the filtering forms on the dashboard with it as well.


tallcoleman:
Thanks @xinweny I like it a lot and would definitely be open to trying it out on the dashboard filter forms later.

One other related thing I wanted to test was making sure that a user can't accidentally double-submit by clicking on the final button twice, since we have some duplicate entries in our dataset. But to test that, we have to set up the submission API to be locally-run when developing. Will submit a separate PR for that and then test for it later.

xinweny:
Thanks for the feedback @tallcoleman! The nice thing about `react-hook-form` is that it has a bunch of useful `formState` variables such as `isSubmitting` or `isSubmitted`. The latter will prevent the submit button from being clicked twice (as long as the `onSubmit` function has not finished running) when paired with the submit button's `disabled` attribute, so hopefully that will prevent future duplicate submissions.

xinweny:
> Thanks @xinweny I like it a lot and would definitely be open to trying it out on the dashboard filter forms later.
> 
> One other related thing I wanted to test was making sure that a user can't accidentally double-submit by clicking on the final button twice, since we have some duplicate entries in our dataset. But to test that, we have to set up the submission API to be locally-run when developing. Will submit a separate PR for that and then test for it later.

Thanks for the feedback @tallcoleman! The nice thing about `react-hook-form` is that it has a bunch of useful `formState` variables such as `isSubmitting` or `isSubmitted`. The latter will prevent the submit button from being clicked twice (as long as the `onSubmit` function has not finished running) when paired with the submit button's `disabled` attribute, so hopefully that will prevent future duplicate submissions.

sonalranjit (approved):
Awesome work @xinweny and thanks for updating the test cases as well!

PR #193, opened 2024-10-20
Configured API url via environment variables

tallcoleman:
- added .env* files to bikespace_frontend and configured NextJS to use them
- changed hardcoded API urls to load from environment variables
- added documentation
- also removed serve-frontend command in makefile which was no longer being used


xinweny (commented):
Thanks for setting this up @tallcoleman, it's been on my mind whenever I see the hardcoded urls and I'm glad it's been taken care of.

Only thing I'd suggest is not to commit the production `.env` and add it to `.gitignore` instead - although it doesn't contain any sensitive keys as of now, we might need to add them in the future.

We could also add the public API key for the Thunderforest map into the `.env` files.

sonalranjit (approved):
Awesome @tallcoleman thanks!
Same thing as what @xinweny has mentioned, normally we don't want to commit production related files but since there isn't any secrets recorded in there it fine.

tallcoleman:
@sonalranjit @xinweny so I looked up to see if there was any advice or best practice on committing the .env files and from what I can tell a possible workflow is:

- environment config - can use and commit e.g. `.env.development` or `.env.production`
- secrets - if needed can use and NOT commit `.env.local` or `.env.production.local` or can use whatever other methods the host offers for loading secrets

If this seems like a fine way to do things I can update the README and also add a comment in the committed `.env` files warning users not to commit secrets?

sonalranjit:
> @sonalranjit @xinweny so I looked up to see if there was any advice or best practice on committing the .env files and from what I can tell a possible workflow is:
> 
> * environment config - can use and commit e.g. `.env.development` or `.env.production`
> * secrets - if needed can use and NOT commit `.env.local` or `.env.production.local` or can use whatever other methods the host offers for loading secrets
> 
> If this seems like a fine way to do things I can update the README and also add a comment in the committed `.env` files warning users not to commit secrets?

Yeah i think it's good for now, yeah a note on the README would be nice but since we all go throught a PR we can be vigilan t on those things.

tallcoleman:
@sonalranjit @xinweny I updated the README and the gitignore; also added top level comments in the `.env` file to remind us that they're for config only.

sonalranjit (approved):
Looks good thanks @tallcoleman

PR #194, opened 2024-10-20
Fixed umami ID used on frontend

tallcoleman:
- also disabled umami load on develop
- also fixed github contribute link on new dashboard


xinweny (approved):
Thanks @tallcoleman! Umami tracking should be already disabled on dev environment in the `track-umami-event.ts` file [here](https://github.com/bikespace/bikespace/blob/main/bikespace_frontend/src/utils/track-umami-event.ts), but I suppose there's no harm writing it at the HTML `<script>` level too.

PR #216, opened 2024-12-04
Fix date string formt in formatHtmlDateValue()

### LINKED ISSUE #205: Bug: Current date picker doesn't seem to be working
Date filter with Custom range, doesn't update the selected date, no mater how many times it gets clicked.
Console log shows ```The specified value "11-01-2024" does not conform to the required format, "yyyy-MM-dd".``` Perhaps it's the date format issue
Image

Image

sonalranjit:
- Fixed the custom date range not updating when selected
- Format date string format to use ISO format when `formatHtmlDateValue` is being called

closes #205

### DIFF
--- a/bikespace_frontend/src/components/dashboard/filter-date-range-custom/FilterDateRangeCustom.tsx
+++ b/bikespace_frontend/src/components/dashboard/filter-date-range-custom/FilterDateRangeCustom.tsx
@@ -118,10 +118,6 @@ export function FilterDateRangeCustom() {
 }
 
 const formatHtmlDateValue = (date: Date) => {
-  return date
-    .toLocaleDateString()
-    .replace(/\//g, '-')
-    .split('-')
-    .map(str => (str.length === 1 ? `0${str}` : str))
-    .join('-');
+  const formattedDate = date.toISOString().split('T')[0];
+  return formattedDate;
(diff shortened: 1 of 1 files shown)


tallcoleman (approved):
lgtm thanks!

kevinluo201:
@sonalranjit @tallcoleman Be careful that `date.toISOString()` will use UTC time. If the parameter `date` has timezone, for example, it is in EST. Then After 19:00 it will be a different date.
CleanShot 2024-12-03 at 21 02 25@2x

kevinluo201:
You can just compose the date by yourselves, like https://github.com/bikespace/bikespace/blob/main/bikespace_frontend/src/components/submission/time/Time.tsx#L16

sonalranjit:
ooo thanks @kevinluo201! I totally didn't consider that case, will push a fix to generate a date string like the one you linked.

tallcoleman:
Might be able to use the luxon library which we're already using in other parts of the front-end. If you parse the date into a Luxon DateTime object, then it looks like you can just call `.toISODate()` and it'll give you the YYYY-MM-DD format while still being time-zone aware.

https://moment.github.io/luxon/api-docs/index.html#datetimetoisodate

PR #218, opened 2024-12-11
Develop new map

xinweny:
Migrate maps from `react-leaflet` to `maplibre`.


tallcoleman:
Thanks @xinweny ! I probably won't have time to look through this until the weekend but look forward to trying it out then.

tallcoleman:
Hi @xinweny I took a look this morning and I think we should try doing the maplibre switch for just the submission page as the first PR and then do a second bigger PR for the dashboard. We could probably keep this branch as-is for the dashboard PR and start a new one for adding maplibre in to the submission page? Let me know what you think.

xinweny:
Hi @tallcoleman, thanks so much for the review. The map component for the submission component is a lot smaller and much less code, so would it be better instead just to do it all in one PR here?

Since it's not ready for merging yet, if you are available to discuss what needs to be done regarding the clustering points, do let me know anytime. Although it might get hard to keep up with changes with `main` especially with the new issues, so if we plan to do merge much later, I will probably just redo this branch at that point.

tallcoleman:
Hi @xinweny sure, we can try and do it all in one PR. I think for the highly-clustered points, we can find a "good-enough" solution and then see if there's opportunity to do some UX research later on what the ideal solution would be for our users.

Spiderification might still work, and would be simplest, though the options available don't seem super robust. Some ones I came across:

- https://github.com/bewithjonam/mapboxgl-spiderifier (uses an overlay over the canvas)
  - https://github.com/ohrie/maplibre-spiderfier - fork of above, might be worth checking out if the one above doesn't work with maplibre out of the box
- https://github.com/nazka/map-gl-js-spiderfy - this one does the spiderification in the canvas, though there's currently [an open issue about react-map-gl compatibility](https://github.com/nazka/map-gl-js-spiderfy/issues/25)
- https://github.com/thuanmb/react-mapbox-gl-spiderifier - this one might work better with react?
- blog post about doing it without using dependencies: https://medium.com/@tomytubert/mapbox-spiderfy-clusters-without-library-8936d4f45347

The other option would be to show a listing of results within the click's "splash radius", which is what the [proof of concept parking map](https://demo-map-app.new-parking-map.pages.dev/) does. This is quite a lot of work though, since we would probably have to:

- Change the click interaction so that it returns a list of matching points using [`queryRenderedFeatures`](https://maplibre.org/maplibre-gl-js/docs/API/classes/Map/#queryrenderedfeatures)
- Control both the matching features and the "selected" matching feature in state 
- Integrate with the clustering so that when a click unclusters and when it selects make sense to the user (probably based on zoom level threshold and whether the matching points includes one or more clusters)
- Build a way to show the matching features in the sidebar component
- If there is one matching feature, select it right away
- If there are more than one matching features, allow the user to select through them using the sidebar component

bikespace_frontend/src/app/(landing)/about/page.mdx
tallcoleman, line 1: I think we can remove this from this PR and keep consistent with main

bikespace_frontend/src/app/(landing)/page.mdx
tallcoleman, line 1: I think we can remove this from this PR and keep consistent with main

bikespace_frontend/src/app/(landing)/style-test/page.mdx
tallcoleman, line 1: I think we can remove this from this PR and keep consistent with main

bikespace_frontend/src/components/submission/submission-form/SubmissionForm.tsx
```diff
@@ -65,21 +64,6 @@ export function SubmissionForm() {
...
-          longitude: position.coords.longitude,
-        });
-        setLocationLoaded(true);
-      },
-      () => {
-        setLocationLoaded(true);
-      }
-    );
-  }, []);
-
```
tallcoleman: I think we should actually keep this useEffect to geolocate the user (without the locationLoaded updates). If we rely on the map geolocate control, it always starts from city hall and then animates moving to the user's location. Whereas this method loads the map where the user is right away. The only downside is that it's a bit confusing to ask for location permissions on page load instead of when this question loads, but maybe there's a way to fix that later.

bikespace_frontend/src/components/submission/location/Location.tsx
```diff
@@ -1,41 +1,68 @@
...
+  const geoControlRef =
+    useRef<maplibregl.GeolocateControl>() as RefObject<maplibregl.GeolocateControl>;
+
+  const {latitude, longitude} = location;
+
+  useEffect(() => {
+    if (!geoControlRef.current) return;
+
+    geoControlRef.current.trigger();
+  }, [geoControlRef.current]);
```
tallcoleman: See comment on `SubmissionForm.tsx` - I think it's cleaner to load the user's location before rendering this component instead of using the geolocate control to do it when the map loads.

One other thing I noticed when I was testing it was that the geolocate control didn't trigger until I clicked the map (which is weird, because this is how the react-map-gl docs say to do it). Putting `    geoControlRef.current.trigger();` in the `onLoad` method of the `MapGL` component does work as intended though.

bikespace_frontend/src/components/submission/location/Location.tsx
```diff
@@ -1,41 +1,68 @@
...
-        <MapContainer
-          center={position}
-          zoom={18}
-          scrollWheelZoom={false}
-          style={{height: '100%'}}
+        <MapGL
+          initialViewState={{
+            latitude,
+            longitude,
+            zoom: 16,
```
tallcoleman: Might be best to keep zoom at 18, zoom 16 is a little too zoomed out for the scale at which someone would be clicking a bike parking location.

bikespace_frontend/src/components/submission/location/Location.tsx
```diff
@@ -1,41 +1,68 @@
...
+              latitude: e.lngLat.lat,
+              longitude: e.lngLat.lng,
+            });
+          }}
         >
-          <TileLayer
-            attribution='&copy; <a href="https://www.openstreetmap.org/copyright">OpenStreetMap</a> contributors'
-            url="https://{s}.basemaps.cartocdn.com/light_all/{z}/{x}/{y}{r}.png"
+          <AttributionControl customAttribution='&copy; <a href="https://www.openstreetmap.org/copyright">OpenStreetMap</a> contributors' />
+          <GeolocateControl
```
tallcoleman, line 55: I think it actually fills the attribution control correctly on its own without having to do the override, so we may not need this part.

bikespace_frontend/src/components/submission/location/Location.tsx
```diff
@@ -1,41 +1,68 @@
...
+            ref={geoControlRef}
+            position="top-left"
+            trackUserLocation
```
tallcoleman: When I was testing it, with `trackUserLocation` enabled, the geolocate control didn't seem to work when the user clicks on it. Maybe it needs `onTrackUserLocationStart` and `onTrackUserLocationEnd` to be implemented if this option is used? User click seems to work without `trackUserLocation` enabled, however.

tallcoleman:
@xinweny FYI I looked at the code for the `/submissions` route and left some comments. Will look at the dashboard code next. Really like how you were able to get rid of some of the extra component layers that we had before!

xinweny:
> @xinweny FYI I looked at the code for the `/submissions` route and left some comments. Will look at the dashboard code next. Really like how you were able to get rid of some of the extra component layers that we had before!

Thank you so much for the really helpful comments (and for the spiderify libraries research) @tallcoleman! I've made the changes as indicated, and I'll look into implementing the spiderification next.

PR #221, opened 2024-12-23
Date filter custom range

### LINKED ISSUE #214: Update date filter custom range to start with today's date and make fully reactive
The custom range option for date filter sets the start date to be the date of the earliest reported issue.
Many users have mentioned it's not ideal to have to start the calendar UI at 2015 and have the user click 10 years ahead to get the current date. 
Ideally the start and end date should be the current date, and the date range should be reactive based on what is clicked.
Example of this behaviour would be the flight date picker on google flights

xinweny:
Fixes #214


tallcoleman:
Hi @xinweny re: the two different API env variables, the purpose is to allow someone to develop the dashboard using the live API while limiting the submission API to the local one only. (The test environment only uses the local API though.) 

I think there's currently an error in the set-up: line 10 of `bikespace_frontend/src/hooks/use-submissions-query.ts` should use the dashboard env variable.

If we wanted to use the local API for both development purposes, we could enhance the local database seeding script to include some of the notable scenarios/patterns in the real data (e.g. highly clustered data points, variations in the parking_time datetime value format, occasional cases from early app uses where the same issue was selected multiple times).

bikespace_frontend/src/components/dashboard/filter-date-range-custom/FilterDateRangeCustom.tsx
```diff
@@ -84,11 +86,7 @@ export function FilterDateRangeCustom() {
           type="date"
           id="filter-start-date"
           name="startDate"
-          value={
-            selectedDateRange.from
-              ? formatHtmlDateValue(selectedDateRange.from)
-              : isoFirst || ''
-          }
+          value={formatHtmlDateValue(selectedDateRange.from || today)}
           min={isoFirst!}
```
tallcoleman: I was thinking for this and the other `input=date` element, we can probably remove the min and max values. 

After writing some tests for this component, I've realized that these built-in validation fields aren't testable and often only work in current implementations when the user clicks through the UI instead of typing.

More importantly, the user might want to define a date period that includes a future date, even if that date is later than the most recent report's `parking_time` value. In fact, the "this year" preset does this already.

tallcoleman (commented):
Hi @xinweny I added in some tests for these components I had written on another branch and updated some of the behaviour where needed. If you think it would be better to use a different implementation, I can update the tests to match.

Let me know what you think about the other comment about the use of the .env variables and then I can approve to merge.

bikespace_frontend/src/components/dashboard/filter-date-range-custom/FilterDateRangeCustom.tsx
xinweny: I agree with this, and eventually we could simplify the form state and validation for the date with the `react-hook-form` and `zod` libraries - I was planning to get this done once PR #192 has been merged (just enabled review, I completely forgot I left it on draft!), as it had made state handling and validation a lot neater and much more manageable on the submission form. But if this happens, the tests would have to be reworked as well.

xinweny:
> Hi @xinweny re: the two different API env variables, the purpose is to allow someone to develop the dashboard using the live API while limiting the submission API to the local one only. (The test environment only uses the local API though.)
> 
> I think there's currently an error in the set-up: line 10 of `bikespace_frontend/src/hooks/use-submissions-query.ts` should use the dashboard env variable.
> 
> If we wanted to use the local API for both development purposes, we could enhance the local database seeding script to include some of the notable scenarios/patterns in the real data (e.g. highly clustered data points, variations in the parking_time datetime value format, occasional cases from early app uses where the same issue was selected multiple times).

Thanks for the explanation @tallcoleman, I think it would be more consistent and straight-forward if we kept one `BIKESPACE_API_URL` in the `.env` file and changed between the local and live API for the `.env.development` and `.env.production` files respectively - this would prevent mixed use of the local and live APIs for different testing purposes and the `.env.development` be used for both development purposes.

As you mentioned, we would have to update the README to enable devs to set the API and database up locally though, as well as for the appropriate seed data, so there would be some additional steps in the development process for the initial setup. This might be better addressed in a new PR though, so for now, I'm happy to revert it back to what it was previously in the meantime.

bikespace_frontend/src/components/dashboard/filter-date-range-custom/FilterDateRangeCustom.tsx
tallcoleman: Very excited to simplify the form logic! Will look forward to reviewing that PR as well. Have removed the min/max fields (and upstream imports they used).

tallcoleman:
> This might be better addressed in a new PR though, so for now, I'm happy to revert it back to what it was previously in the meantime.

Sounds good, have reverted for now and we can open a new issue to completely remove the live API as a dev dependency.

tallcoleman (approved):
Thank you for fixing this @xinweny !

PR #232, opened 2025-02-04
Readme updates

tallcoleman:
Updated the main README to align with the URL changes and added anchor link generation to mdx to enable a change for the "get involed" link


sonalranjit (approved):
Thanks @tallcoleman !

PR #234, opened 2025-02-07
Move old landing page and dashboard src code into archive directory

### LINKED ISSUE #233: Archive old landing page and dashboard
Since we have successfully migrated the landing page and dashboard to react, we no longer require to maintain the old landing page and dashboard. 
Ideally we can just remove it from the repo, since the history is tracked it will always be available by just rolling back to a commit.
But also we can just create a folder called `archive` and place them there as well.

sonalranjit:
Closes #233 

- Moved old landing page and dashboard src into `archive`
- Created a `index.html` for cloudflare page to redirect app.bikespace.ca to bikespace.ca


tallcoleman (approved):
Thanks @sonalranjit ! 

FYI I updated the main README and removed the old dashboard stuff from the makefile as well.

PR #235, opened 2025-02-13
Bugfix - frontend uses wrong submission post format

tallcoleman:
Issue: front end does not provide data to the /submissions [POST] route in the expected format.

To replicate: run the frontend and API locally and try making a submission.

Fix provided here: add a converter function between `SubmissionSchema` and `SubmissionPayload`. Also fixed hard-coding of API url.

### DIFF
--- a/.github/workflows/test-preview.yml
+++ b/.github/workflows/test-preview.yml
@@ -42,6 +42,8 @@ jobs:
         run: make run-pytest
       - name: Coveralls
         uses: coverallsapp/github-action@v2
+        with: 
+          fail-on-error: false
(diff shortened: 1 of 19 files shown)


xinweny:
Thanks @tallcoleman for the fix - apologies for missing it. Seems like the API tests are failing though, but I think it's because [coveralls](https://coveralls.io/) is down for maintenance so we can't merge it yet. We could either wait it out or update the CI build to allow the coverall step to fail on error as described [here](https://status.coveralls.io/?utm_source=embed).

xinweny (approved):

PR #239, opened 2025-03-09
Bump up buildpack python version to supported 3.12.8 version

sonalranjit:
Python 3.12.4 isn't supported on the paketo buildpacks anymore.
Bumping python version to 3.12.8 for fly.io


tallcoleman (approved):
🐍

Issue #213, opened 2024-11-30
Improve sidebar drawer toggle on mobile


sonalranjit:
User feedback:
"In the mobile/small screen mode, the horizontal line ui element to make the filers/results appear wasn’t initial clear to me as a button (I kept on trying to drag in up with the thumb). There might be a clearer way to emphasize that is a button to open/close."

Suggested implementation - use up/down chevron icon

Also the when the drawer is close, the map doesn't rerender the space the drawer was initially using.

okpora:
Mockups are available for review at the following link:

https://www.figma.com/design/THv9hOFLFmmhrJMDyAF0SN/BikeSpace-Draft?node-id=2450-1790&t=vb9Tc6f7CL4fkDdg-1

portolosalexandra:
Long Term Solution for Mobile:

This solution requires us to add a Title/Header to each tab. Additionally, Feed tab is an exception that expands halfway instead of full screen on mobile. 

Desktop:

<img width="400" alt="Image" src="https://github.com/user-attachments/assets/b9141cf2-48aa-449e-89e2-0e083a281cc4" />

Mobile Menu (Default):

<img width="267" alt="Image" src="https://github.com/user-attachments/assets/408f265d-e4c3-48ae-ba47-38ff328b2e7d" />

Mobile Menu (On-Click - Interaction Design):

<img width="270" alt="Image" src="https://github.com/user-attachments/assets/f12ecce1-c098-4063-bfe1-4237bb7aa2d7" />

Mobile Menu (Expanded):

<img width="263" alt="Image" src="https://github.com/user-attachments/assets/fb33aa9e-dc55-4ff7-98ac-1674d2587fca" />

Mobile Insights Tab:

<img width="271" alt="Image" src="https://github.com/user-attachments/assets/b96a1695-8ed9-48fb-9070-c375f7dab489" />

Mobile Insights Tab - Back Button on Click:

<img width="268" alt="Image" src="https://github.com/user-attachments/assets/5426ff4e-ef6f-4ebe-87cd-2d41a7d8fb37" />

Mobile Feed Tab:

<img width="263" alt="Image" src="https://github.com/user-attachments/assets/e756ed9d-bd73-4437-9572-0f55de006c71" />

tallcoleman:
@okpora have coded this up - preview link here: https://638643e6.bikespace-v2.pages.dev/

tallcoleman:
Closed by #240

PR #240, opened 2025-03-09
Update sidebar toggle

tallcoleman:
Implements #213

Have added the icon and the styling, ~~just needs a hover animation and some testing~~ (done)


tallcoleman:
@xinweny I added a change to this to invalidate the map size when the sidebar toggles so that it fills the whole screen via commit [b5a1d3b](https://github.com/bikespace/bikespace/pull/240/commits/b5a1d3b12bb525fee304dc4a112649a32486e278). To do it, I lifted the sidebar open/close state up so that the map can watch it and run `invalidateSize` when it opens/closes. But if you can think of a simpler/cleaner way to do it, let me know?

xinweny:
I've moved the sidebar state into the global store instead (renaming it to `useStore` instead of `useSubmissionsStore` to be more general), creating another slice `UiSlice` where we can put our UI states for the dashboard. This would prevent prop drilling - alternatively, we could also use plain React Contexts but since we are already using `zustand`, we might as well put the UI state together with the other app states.

Do check out commit [b666f02](https://github.com/bikespace/bikespace/pull/240/commits/b666f0294f5ea44c298ee435fc57d52de2f14131) whenever you have the time and let me know what you think!

sonalranjit (approved):
🤘🏽

PR #244, opened 2025-03-26
API add received datetime

tallcoleman:
Submission Received DateTime

What problem does this solve? 

Data quality — the datetime that issues are reported is currently ambiguous:
- parking_time represents the datetime that the user encountered the bicycle parking issue, and is arbitrarily editable by the user when making a report
- id corresponds to the order that reports were received for #1242 onwards, but does not appear to consistently relate to this order for past entries:
  - ids 1–41 have parking_time ranging from August to October 2023
  - ids 42–1019 have parking_time increasing from to June 2018 to June 2019, but with many entries that are farther in the future compared to the dominant sequence
  - ids 1020–1241 have parking_time ranging from July to August 2018 and a couple values (especially at the end) in 2019 and 2020

Proposed solution:

- Add a submitted_datetime field to the bikeparking_submissions table which is added as part of the logic for the /submissions [POST] route
- For past submissions, submitted_datetime will be null

TODO

- [ ] document the implementation date that this field was added (through release function?)
- [x] add this as an output to the [GET] route
- [x] document workflow for adding and testing database migrations
- [x] add to pytest suite
- [x] update `bikespace-open-api.yaml` for the public api documentation
- [x] document how to update the open api yaml
- [x] add a basic way to see the submitted_datetime values in the frontend
- [x] add unit tests for frontend components (was able to do `FeedSubmissionItem` component but not `MapPopup`


tallcoleman:
Resolved: ~~Something that would be helpful to get advice on: `submitted_datetime` is nullable (for grandfathering reasons), but I'm not sure how to test the scenario where that column is null in pytest since making a submission doesn't result in a null value for that column.~~

tallcoleman:
@sonalranjit is there a preferred way to update `bikespace-open-api.yaml` (e.g. cli tool) or is it hand-updated?

sonalranjit:
@tallcoleman no special tool to update the yaml. Just have to be mindful of the strict yaml format

tallcoleman:
@sonalranjit this is ready for review. 

## Key notes for consideration:

- Let me know if you think `submitted_datetime` is a good name for the added column
- How should we document the minor change in the API? I bumped the version in the open api YAML for the api to `2.0.1`, but should we plan on doing a repo-wide release as well? (which would be release `2.0.2`) Should the version numbers be the same for both?

## Summary of changes:

### Main Changes:

- Added `submitted_datetime` column to `bikeparking_submissions` table to indicate the time that the server received a report. This value will be null for reports received before this API change. Unlike `parking_time`, it returns a timezone-aware value.
- Updated database seeding script and unit tests for python - tests both regular (date) and null values for `submitted_datetime`
- Generated and tested database migration scripts
- Updated OpenAPI documentation to include `submitted_datetime`; also made a clearer distinction between POST parameters and GET parameters
- Added option to view submitted_datetime in frontend dashboard by hovering over date values in feed or map pop-up (plus unit test for feed item component)
- Added `downgrade-db` make target (used for migration testing)
- Added documentation on updating OpenAPI yaml and how to add database migrations

### Incidental Changes:

- updated `run-pytest-terminal` make target to show lines where test coverage is missing
- The response value for /submissions POST requests now include the submission_id. This would enable the post-submission page to link a user directly to the dashboard after submission (see: https://github.com/orgs/bikespace/projects/1?pane=issue&itemId=65627143 )

### Screenshots of frontend implementation

Map Popup - null `submitted_datetime` value (mouse is hovering over date description)
Screenshot 2025-03-27 at 6 29 10 PM

Feed item: regular `submitted_datetime` value (mouse is hovering over date heading)
Screenshot 2025-03-27 at 6 29 42 PM

sonalranjit (approved):
Awesome thanks @tallcoleman , feel free to squash and merge it

PR #246, opened 2025-03-30
Remove line that creates migrations in fly_release.sh

sonalranjit:
- Remove python db migrate line in fly release shell script


tallcoleman:
@sonalranjit I added a step in the `test-preview` and `deploy-to-fly` actions to check if a database migration needs to be generated. This should prevent people from pushing to main if they forget to generate the necessary migration scripts for schema changes.

tallcoleman (approved):
Thanks @sonalranjit , feel free to merge once you've reviewed the added CI step to check if database migrations are needed

PR #247, opened 2025-03-31
Remove db stamps head from fly_release.sh

sonalranjit:
- Remove db stamp heads command before applying the migrations


tallcoleman (approved):
thank you!!!

Issue #209, opened 2024-11-23
Create an instruction modal or popup guiding users how to use the dashboard


sonalranjit:
Various users have mentioned they've had to guess the functionalities of the dashboard. There are no instructions on how the filtering works or any explanation of the data that is being visualized. 
A simple popup modal or a separate page listing how to use the dashboard would help users not feel lost.

sonalranjit:
Needs a design component

portolosalexandra:
Hey everyone, please see the updated design for this ticket - thanks for all the feedback @okpora @tallcoleman @xinweny and @sonalranjit ... a bit later on I will share a modified tentative design that also includes a 'launch tutorial' option

<img width="476" alt="Image" src="https://github.com/user-attachments/assets/ded96d34-e692-4398-a643-829427deb6b9" />

You can find the Figma design file here:
https://www.figma.com/design/GN8BdmaWZeWRs0qMW5eccVvk/BikeSpace-dashboard-mockup?node-id=785-2287&p=f&t=uitd9Oft8rfsotTa-0

portolosalexandra:
Please also see the mobile design:

<img width="257" alt="Image" src="https://github.com/user-attachments/assets/aef11c9e-56ee-4681-8406-898bf6b8a3c1" />

Expanded menu:

<img width="263" alt="Image" src="https://github.com/user-attachments/assets/f257c028-50b1-4ef4-b2d4-11ade67f1b67" />

And interaction for 'on click':

<img width="265" alt="Image" src="https://github.com/user-attachments/assets/b5a3d648-f927-41c1-a36c-2080918a587b" />

Info page:

<img width="320" alt="Image" src="https://github.com/user-attachments/assets/6f976bc0-f6e6-4e0d-af9e-ca22ddad671d" />

tallcoleman:
Closed by #249

PR #249, opened 2025-04-13
Sidebar info view

iainmcampbell:
- https://github.com/bikespace/bikespace/issues/209

- Implemented redesigned sidebar tabs
- Added info panel

<table>
<tr><td colspan="3">
Screenshot 2025-04-16 at 10 31 59 PM
</td></tr><tr><td colspan="3">
Screenshot 2025-04-16 at 9 56 55 PM
</td></tr><tr><td>
Screenshot 2025-04-16 at 9 57 06 PM
</td><td>
Screenshot 2025-04-16 at 10 32 18 PM
</td><td>
Screenshot 2025-04-16 at 9 57 36 PM
</td></tr></table>

I’ve tried to refactor the sidebar styles to be mobile-first.

I used SVGR (svg as jsx file) for the sidebar icons so we can style them using CSS.

Since the server doesn't know whether we're mobile or not, this new pattern where the sidebar content is open by default on desktop and closed by default on mobile is a bit of a challenge to implement using server-side rendering. I made it work by defaulting the sidebar content to closed, so we never render it on the server. Once the client is hydrated, we can use a media query to know we're in a desktop viewport, and then render the sidebar content. Let me know if this makes sense, or if there’s a better way to handle this.

I tried to use the colour variables in [variables.module.scss](https://github.com/bikespace/bikespace/blob/main/bikespace_frontend/src/styles/variables.module.scss), they might not match the Figma 100% but we should be sticking to a standard colour palette, right?

I tried to use the existing SidebarButton component, which is close to but doesn't quite match the mockups.

## Questions

Where is the data guide PDF?

I assume the 'Data last updated' in the info view is meant to represent the date of the most recent submission? I tried to piggyback off the Feeds view, but that data isn't always available. I can make a new query - does the backend have a way to fetch only the most recent submission? This date is also available in the Feed view, so I don't know if it’s worth setting up a new endpoint just for this.


tallcoleman:
Thanks for taking this on @iainmcampbell ! 

Just wanted to check since I don't understand this note:

> having the sidebar content open by default on desktop and closed by default on mobile is a bit of a challenge, the server doesn't know whether we're mobile or not

Which server needs to know the client state? The way we deploy is to build the site to static so there's no server used for frontend content, only the 'backend' server to get the data from the API.

iainmcampbell:
> Which server needs to know the client state? The way we deploy is to build the site to static so there's no server used for frontend content, only the 'backend' server to get the data from the API.

Ah, gotcha. I was getting SSR errors about media queries, I assume because next.js does SSR in dev mode. but a static build would have the same issue - it _very_ definitely cannot know anything about the client’s computer, eg whether they're on a phone or not.

These mockups have client state working differently in mobile vs desktop. Previously the mobile view just had different CSS, but now in mobile we have different state too. Now we don't render any sidebar tab content by default, and introduce a new concept of "opening" a sidebar tab, which we don't have on desktop. I got around this by not opening a sidebar tab by default, and then once the page is rendered, check if we're on desktop, and then open the panel.

I wanted to mention it in case there’s a more elegant way to solve this problem.

iainmcampbell:
ok I just tried the classic "is it SSR" test of loading bikespace.ca with JS disabled, and like you said the static build doesn't do server-side rendering in production. it's just `<body><script></body>`.

So it's puzzling that it's trying to do it in development. if we can turn off SSR in dev, we could remove a bunch of future complexity.

I don't have a lot of experience with next.js so I may be misunderstanding here?

sonalranjit:
@iainmcampbell to build the site statically on your local environment i believe you can run ```npm build``` or ```npm start``` i believe but im not a 100% confident.
If you run it in develop mode with hot reloading i believe it runs with SSR.
Also im not sure if you saw the preview link, sorry it's not too obvious in the checks, your PR change are deployed here https://48cd184a.bikespace-v2.pages.dev/ 
Link is hidden underneath the Cloudflare Pages block.

bikespace_frontend/src/assets/images/data-guide.png
tallcoleman, line 1: Do we have the svg versions of these saved somewhere?

bikespace_frontend/src/components/dashboard/sidebar-tab-content/_SidebarContentInfo.tsx
```diff
@@ -0,0 +1,44 @@
...
+import dataGuide from '@/assets/images/data-guide.png';
+
+import styles from './_SidebarContent.module.scss';
+import {SidebarLinkButton} from '../sidebar-button';
+
+const content = {
+  lastUpdated: 'Data Last Updated',
+  about: 'About the Dashboard',
+  aboutContent:
+    'The BikeSpace Dashboard is based on issues reported through the publicly accessible Bikespace Issue Reporting web app.\nLaunched in July 2018, BikeSpace allows cyclists to self-report bike parking concerns they come across in the Toronto, Canada area.',
```
tallcoleman: What's the benefit of putting the text content here instead of just writing it into the return function below?

Also, the `\n` in `aboutContent` does not get rendered FYI. You could either make it into a list and use `map` to render each value as its own paragraph or put the content into an MDX file that gets imported instead. (When I tested this, in `tsconfig.json`, I had to add `"mdx"` to `types` and `"**/*.mdx"` to `include`).

bikespace_frontend/src/components/dashboard/sidebar-tabs/icons/export.tsx
tallcoleman, line 1: An a11y challenge with using these svgs on their own (instead of as the src for an img tag) is that you can't set `alt=""` to indicate that the image is decorative. This got flagged when I tested with the Firefox accessibility tools. The equivalent options for svg don't seem to have a way to indicate the icon is decorative either.

One option would be to add `role="img"` in the top-level `svg` tag and a title tag (e.g. `<title>Export</title>` before the `path` tag). In theory this might mean a screen reader reads the nav title twice (once via the svg title tag, and once via the label on the button element). Though when I tested it with VoiceOver, it just read the button label.

tallcoleman (changes requested):
Thanks @iainmcampbell this is really great! Very exciting to have this implemented.

I'll note the general comments I jotted down during testing; if you want me to help with any of these, just let me know and I'll push some commits to the branch.

- The Info, Insights, and Filters tabs did not scroll for me on desktop when the content exceeded the viewport height.
- The line spacing for the text on the Info tab (170%?) seems quite generous compared to other body text (e.g. comment text on the Feed tab; text on the About page, which is 140%)
- BikeSpace should be property capitalized everywhere it is used
- I pushed a commit to add the data dictionary PDF to the /public folder - please add link to this PDF in the button on the Info tab
- The Feed tab serves to show pop-up text on mobile, so it should actually be half-height on mobile when opened (see screenshot from the Figma design below)
- The width of the dashboard nav is _slightly_ different when Feed/Info are selected s. when Insights/Filters are selected.

There is also a bug where the dashboard navigation stops working if a map pin is selected. This appears to be an existing bug, so if you don't want to address that in this PR, I'll open an issue and fix it after this is merged.

Thanks again and let me know if you have any questions!

Screenshot 2025-04-18 at 8 49 53 AM

tallcoleman:
Sorry, one more thing I forgot - similar to the mobile styling when a tab is open, there needs to be padding below the dashboard nav on mobile so that the buttons don't conflict with the OS UI elements. e.g. see screenshot below on an iPhone running the site as a PWA.

bikespace_frontend/src/components/dashboard/sidebar-tabs/icons/export.tsx
tallcoleman: Update: I looked into this some more:

- [The solution](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/img_role#svg_and_roleimg) for a non-presentational svg is to add `role="img"` and an `aria-label` to the svg. In this case though, I think the right solution would be to add `role="presentation"`, which is equivalent to an img tag with `alt=""`.
- The reason that VoiceOver isn't reading the children of `button` is that the [ARIA button role treats all descendents as presentational](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/button_role#all_descendants_are_presentational). So the Firefox accessibility tools are probably flagging it incorrectly, but adding `role="presentation"` should fix it just in case other screen readers or browsers handle `button` slightly differently.

bikespace_frontend/src/assets/images/data-guide.png
iainmcampbell: I can use SVGs for these instead if that would be better, I just exported them from Figma

bikespace_frontend/src/components/dashboard/sidebar-tab-content/_SidebarContentInfo.tsx
iainmcampbell: old habits. if you're translating your site, you generally have lint rules to prevent strings in the JSX. if that's not a concern here, then I’ll just put the strings in the JSX, then no need to jump through hoops to get newlines to render.

iainmcampbell:
thanks for a very thorough review!!

not done yet, but I believe I've addressed the following:

> - The Info, Insights, and Filters tabs did not scroll for me on desktop when the content exceeded the viewport height.
> - The line spacing for the text on the Info tab (170%?) seems quite generous compared to other body text (e.g. comment text on the Feed tab; text on the About page, which is 140%)
> - BikeSpace should be property capitalized everywhere it is used
> - I pushed a commit to add the data dictionary PDF to the /public folder - please add link to this PDF in the button on the Info tab
> - The Feed tab serves to show pop-up text on mobile, so it should actually be half-height on mobile when opened (see screenshot from the Figma design below)
> - The width of the dashboard nav is slightly different when Feed/Info are selected s. when Insights/Filters are selected.

bikespace_frontend/src/assets/images/data-guide.png
tallcoleman: OK thanks!

bikespace_frontend/src/components/dashboard/sidebar-tab-content/_SidebarContentInfo.tsx
tallcoleman: Right - translation! makes total sense. I think in this case, strings in the JSX is fine.

tallcoleman:
> not done yet, but I believe I've addressed the following:

Thanks @iainmcampbell , appreciate it! Let us know when you want someone to take a final look.

iainmcampbell:
Ok, I think I got everything:

- added padding below mobile dashboard nav
- added `role="presentation"` to the SVG icons `<svg>` elements
- converted the info page images to SVGs

should also note I loaded the Sidebar component dynamically using Next.js’s `dynamic()` function, which seems to be the least hack-y way to avoid SSR. This allowed me to remove a bunch of logic (https://github.com/bikespace/bikespace/pull/249/commits/56190eab28b82113bb5cd0be991f387504b077b0)

tallcoleman:
@iainmcampbell only one bug to fix and then this is ready to go:

For some reason on Safari, when the sidebar content exceeds the available height and has to scroll, the button disappears (examples with different heights below):

Screenshot 2025-05-03 at 9 21 10 AM

Screenshot 2025-05-03 at 9 22 30 AM

I've been unable to puzzle this one out, so if you have any ideas, let me know!

I was able to fix another bug I've run into before with using `vh` on Safari and made a couple other very small style tweaks.

tallcoleman (approved):
@iainmcampbell thanks again for all your help on this! I think I fixed the bug with the back to map button not showing on Safari. Take a look and if it all seems OK with you, feel free to merge.

iainmcampbell:
@tallcoleman nice catch, fix looks great!

PR #257, opened 2025-05-18
Add address search to parking map

tallcoleman:
Adds a search box to the /parking-map page that allows the user to search for a location by address or place name.

Address search is done using [photon](https://photon.komoot.io/). Search requests are tuned to the user's current map centre and limited to results within the bounding box for the City of Toronto. Typing in a search query will automatically start showing results; the user can then select one of the results to centre the map on that location. Search results can be cleared using the 'x' button.


bikespace_frontend/src/utils/map-utils/geocoder-search/GeocoderSearch.tsx
```diff
@@ -0,0 +1,218 @@
...
+  }, [selectedResult]);
+
+  // Get location search results using the photon API
+  // Prioritize map centre and limit to Toronto bounding box
+  const query = useGeocoderQuery(
+    debouncedInputValue,
+    map ? map.getCenter() : undefined,
+    bbox,
+    resultsLimit
+  );
```
iainmcampbell, line 197: it would be cool to cache the last search result so we can pull it up again instantly. I'm noticing when the results are hidden (click on a parking feature, or the bikespace tab becomes inactive) and then I go back to the search results, we call the komoot api again even though we technically already have the results.

bikespace_frontend/src/utils/map-utils/geocoder-search/GeocoderSearch.tsx
```diff
@@ -0,0 +1,218 @@
...
+  // minimize search results when parking features are selected
+  useEffect(() => {
+    if (!isMinimized || !selectedResult) return;
+
+    const address = getAddress(selectedResult);
+    const name = selectedResult?.properties?.name;
+    setInputValue((name ? `${name}, ` : '') + address);
+  }, [isMinimized]);
+
+  function SearchResults() {
```
iainmcampbell: it’s a bit odd to define a React component inside another React component. would be cleaner to pass `SearchResults` the props it needs, and define it as its own top-level function.

iainmcampbell (commented):
excellent feature! I especially like the loading state, it’s fun and I love that’s implemented with CSS.

Should we add a close button to the selected parking feature, so you can get back to the search results? You can click on the map, but that’s implicit, there’s no visible affordance. I don’t think we really had this issue before because the only thing to go back to was the legend, which has a visible affordance (collapsed state). maybe we should collapse the search results too, instead of hiding them?

bikespace_frontend/src/utils/map-utils/geocoder-search/GeocoderSearch.tsx
tallcoleman: Fixed in https://github.com/bikespace/bikespace/pull/257/commits/c0b2d175958f72f291253c443180b3ddbe6dd79b - let me know what you think

bikespace_frontend/src/utils/map-utils/geocoder-search/GeocoderSearch.tsx
tallcoleman: I changed the behaviour so that the minimize / open actions don't change the query string (so no additional queries are made) - does that work?

tallcoleman:
> Should we add a close button to the selected parking feature, so you can get back to the search results? You can click on the map, but that’s implicit, there’s no visible affordance. I don’t think we really had this issue before because the only thing to go back to was the legend, which has a visible affordance (collapsed state). maybe we should collapse the search results too, instead of hiding them?

I think this is a good suggestion - added the "clear selection" button in https://github.com/bikespace/bikespace/pull/257/commits/7352d45e2ec742cf53fea36b1b92dc8441281955

I changed the way that the location search minimizes and re-ordered the components, which should probably address the other notes you had here

Ultimately my goal with this PR is to get the user interaction good enough, and then we can do additional user testing to refine things and also to figure out the most logical way to add things like filters (e.g. if you're only looking for bike parking that's covered, or good for cargo bikes)

Thanks again for the review!

tallcoleman:
Todos before merge:

- [x] update from main
- [x] fix error with location search crashing the page if the user does not have location enabled and add a test case for this
- [x] changing the location search query text should clear location icons on map

tallcoleman:
@sonalranjit crash on location search if user has geolocation permission off is now fixed!

sonalranjit (approved):
awesome @tallcoleman

PR #263, opened 2025-06-06
Change API deployment to use Docker instead of buildpacks

tallcoleman:
Using Docker to deploy flask apps seems to be the current preferred path for fly deployments. Have updated to deploy the API via Docker and have tested successfully by deploying to fly separately.

**⚠️ I would recommend pulling a backup of the database before merging this into production** (just in case?)

Major changes:

- fly deployment now done using Dockerfile instead of buildpack and Procfile
- local running of the API for development is now done using Docker compose
- production config converts "postgres://" prefix to "postgresql://" since SQLAlchemy only supports "postgresql://"
- fixed issue with order that migrations were applied, allowing for removal of table creation code in recreate_db command
- updated documentation

Incidental changes:

- updated make target naming for consistency and removed unused make targets
- added messages to migration docstrings

Unsolved challenges:

- I don't know how to run pytest using Docker, so the python tests are still running as before (only using container for the test database, not for the API)
- Workflow with making API changes may need to be tested further

References consulted:

- https://docs.docker.com/guides/python/
- https://fly.io/docs/python/frameworks/flask/


sonalranjit:
whoa awesome @tallcoleman , i just did a quick check of running it locally and it works great!
I'll spend a lil bit more time reviewing it tonight or tomorrow.

sonalranjit:
I took a backup of the current database, i was testing restoring it but i was running into trouble. I'll try to get a successful restore and then we should be good to go.

bikespace_api/bikespace_api/config.py
```diff
@@ -31,4 +31,6 @@ class TestingConfig(BaseConfig):
 class ProductionConfig(BaseConfig):
     """Production configuration"""
 
-    SQLALCHEMY_DATABASE_URI = os.environ.get("DATABASE_URL")
+    SQLALCHEMY_DATABASE_URI = os.environ.get("DATABASE_URL", "").replace(
```
sonalranjit, line 34: At some point this has to be passed on to the flask container. Since the postgres instance is a separate app, i think we might have to declare it in fly.toml

tallcoleman: I think that the deployment instance is connected to the deployed database using the `DATABASE_URL` secret. This is to help in case the postgres url is `postgres://` instead of `postgresql://`, which sometimes causes problems.

sonalranjit (approved):

PR #265, opened 2025-06-08
Ignore archive in vscode’s file search

iainmcampbell:
I keep getting results from the archive and confusing myself when using `cmd` + `p`


tallcoleman (approved):
I had no idea you can do this - great idea!

PR #268, opened 2025-06-15
API tests and makefile cleanup

tallcoleman:
Changes:

- restore database clear needed to ensure API tests can run back to back locally
- inline and update .PHONY marks in makefile

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -81,7 +81,6 @@ test-api: setup-py launch-db db-test-server
 	export TEST_DATABASE_URI=postgresql://postgres:postgres@localhost:5432/bikespace_test && \
 	cd $(BIKESPACE_API_DIR) && \
 	$(PYTHON) $(MANAGE_PY) recreate-db && \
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -42,30 +44,38 @@ $(VENV): check-python-version $(BIKESPACE_API_DIR)/requirements.txt
 	$(PIP) install -r $(BIKESPACE_API_DIR)/requirements.txt
 	touch $(VENV)
 
+.PHONY: pip-freeze
 pip-freeze: $(BIKESPACE_API_DIR)/requirements.txt
 	$(PIP) freeze > $(BIKESPACE_API_DIR)/requirements.txt
```
sonalranjit, line 49: technically this isn't a phony target because it is creating a file `requirements.txt`

tallcoleman: Since the output file isn't named `pip-freeze` would it still make sense to leave this target as phony?

If non phony = makes a committed file I think the API migration command would also count?

sonalranjit: Yeahh actually thinking over it again, i think it's fine to leave it as is.

sonalranjit (approved):
Awesome thanks @tallcoleman

PR #273, opened 2025-06-28
Parking map add filters

tallcoleman:
Adds two new UI elements to the parking map page: a filter control to show or hide features based on the values of one attribute key and a layer control to show or hide the bicycle network background layer. Both are organized within a `<details>` toggle box that is closed by default.

## Filter Control

The filter control allows the user to show or hide features based on selecting or deselecting the possible values from one attribute key. The user can either select or deselect individual values, or use the "select all" and "clear all" buttons. 

By default, the attribute key presented is `bicycle_parking` (type of bike parking, e.g. post and ring, rack, stands, etc.), but the user can select others from a pre-constrained list defined in `parking-map-filter-default-properties.ts`. The pre-constrained list was selected based on values that the average user might find useful and relatively easy to understand, and may be adjusted in the future based on the results of user testing.

The list defined in `parking-map-filter-default-properties.ts` also provides more user-friendly names for attribute keys (e.g. "Bicycle Parking Type" for `bicycle_parking`) and simple data type information that is used to properly sort the values of the attribute key. Most attributes are sorted alphabetically, but "integer" attributes (which usually still contain a small number of text values) show the alpha-sorted text values first, followed by numbers sorted by their values. This prevents the problem of number strings being sorted as, e.g. "1, 10, 2, 21". In the future, these data types could be used to further customize the interface presented to the user (see notes at the end).

Since the `<input>` element expects its attributes to be strings, the component also has a function `transformPropertyOptions` that converts null and boolean values into either a "value" string or a "description" string, depending on the use case. Using this function, a checkbox to select or deselect null values is composed as follows, using the "value" `__undefined__` and the "description" `(not specified)`. The other edge cases handled by the function are `''` (blank string), `true`, and `false`.

```html
<div>
  <input id="__undefined__" type="checkbox" name="__undefined__">
  <label for="__undefined__">(not specified)</label>
</div>
```

Currently, when a user clicks on features on the map, the sidebar components that show feature descriptions get their underlying data directly from the map using the [`queryRenderedFeatures` method](https://maplibre.org/maplibre-gl-js/docs/API/classes/Map/#queryrenderedfeatures). For the filters, this approach (or using [`querySourceFeatures`](https://maplibre.org/maplibre-gl-js/docs/API/classes/Map/#querysourcefeatures)) was not viable, since the data returned by either method only includes features within the current viewport. There does not appear to be a method in MapLibre to return all data from a layer, regardless of whether it is within the viewport. Using the data from the map would therefore mean that the list of possible options would potentially become stale as soon as the user pans or zooms.

To address this, the new `useParkingDataQuery` hook fetches the data directly, so that the list of values to filter on always represents the possibilities present in the entire dataset. This means that the parking map page is querying the same data twice: once via the map's Source component and once via `useParkingDataQuery`. Although a more elegant solution may be possible, there was no noticeable performance impact of this double-fetching during manual testing.

## Layer Control

The layer control shows or hides the bicycle network background layer. When there are very few results from a user's filter, this can make it easier to locate those parking features on the map. In the future, this could be expanded to allow greater control over other layers (e.g. choice between vector basemap or satellite imagery for the background, also showing the issue reports, etc.).

To ensure that the bicycle network layer stays below the bicycle parking layer when it is re-added to the map, the `beforeID` attribute was added for layers in the `BicycleNetworkLayer` component and set to a new `parkingFirstLayerId` constant exported by the `ParkingLayer` component.

## Other Changes

A new `shared-ui` component folder has been created for UI elements that are used in both the dashboard page and the parking map page. The components added include:

- a styled `<button>` element `SidebarButton`
- a styled `<details>` element `SidebarDetailsDisclosure` (and a `SidebarDetailsContent`component to apply consistent styling for the content that goes below the `<summary>` tag)
- a styled `<select>` element `SidebarSelect`

These are functionally the same as their respective HTML elements, but with default styling that can be overridden either with more specific class selectors or the component's style attribute. This helps to keep styling consistent between both pages and allows for these common elements to be quickly re-used in other interfaces.

### Numbers

Properties specified as the "integer" type currently have their number values sort correctly, but it would be more intuitive to have, e.g. a min and max option or some kind of slider UI instead of having a checkbox for every number that appears:

Screenshot 2025-09-21 at 2 46 49 PM

### Making values more user-friendly

The filter options currently just show the "raw" OpenStreetMap tags, which may or may not make sense to the user. For some properties, a mapping should be added, [similar to what is done with bicycle parking types in the selection UI](https://github.com/orgs/bikespace/projects/3?pane=issue&itemId=121648915), pulling from the OSM wiki or iD presets.

Screenshot 2025-09-21 at 2 49 56 PM

Some properties (e.g. `panoramax`, `mapillary` or `wikidata`-type tags would benefit from integration of [tag2link](https://www.npmjs.com/package/tag2link). 

Other tags would benefit from logic to map null values (displayed as "(not specified)" or "(blank)") to a reasonable default, e.g. "no" for the `covered` tag and "yes" for `access` tag.

### Dates

An additional "type" option could be added for date properties, e.g. `check_date`:

Screenshot 2025-09-21 at 2 53 16 PM

Conforming values could be converted to Date values and then filtered in a more intuitive way.

### Property Conversion

Some attributes would make more sense if they were transformed into something more intuitive for the user. E.g. lifecycle tags could be combined and converted into some kind of "status" type property where something like `construction:amenity=bicycle_parking` becomes a filterable value like "Under Construction". In some cases, this may mean combining the values of multiple tags, either in the API or using the frontend code. 

Screenshot 2025-09-21 at 2 54 57 PM

Another good example would be `image` and related tags, where the properties are simplified into whether an image tag is present or not. This might also be a case where several image-related tags should be displayed together, e.g. `image`, `panoramax`, and `mapillary`.

See also: [handle additional image tags](https://github.com/orgs/bikespace/projects/3?pane=issue&itemId=123053012)

### Additional documentation

Some properties, e.g. `fixme` would benefit from a short explanatory blurb to explain the meaning of the tag. (In this case, that it is asking for users to take action by editing OpenStreetMap, not that the real-life parking feature needs fixing). Many of these cases might be handled by adding in the wiki link-outs used by editors like iD or Go Map!!

### Opening Hours

Another possible value type, could be parsed using something like [opening hours package](https://www.npmjs.com/package/opening_hours) and add a more intuitive UI for the user to filter based on whether a feature is open on a specific date/time or "open now".

Screenshot 2025-09-21 at 3 01 22 PM


bikespace_frontend/src/components/parking-map/parking-map-page/parking-map-page.module.scss
```diff
@@ -50,36 +50,15 @@ p.cardBody {
...
-
-  summary {
-    height: 2rem;
-    line-height: calc(2rem - 4px); // center text vertically
-    font-weight: 600;
-    color: $color-primary-d50p;
-    border-bottom: none;
-  }
-
-  &[open] summary {
```
MichaelMifsudSweeney, line 67: Would be great to put cursor:pointer here to make the mouse hover more intuitive

bikespace_frontend/src/components/parking-map/parking-map-page/map-filters/ParkingMapFilter.tsx
```diff
@@ -0,0 +1,295 @@
...
+        } as FilterPropertyAttributes)
+    );
+    return newPropertyListAttributes.toSorted((a, b) =>
+      a.description.localeCompare(b.description, undefined, {
+        sensitivity: 'base',
+      })
+    );
+  }, [features]);
+
+  const propertyOptions: Set<string> = useMemo(() => {
```
MichaelMifsudSweeney, line 174: I wonder if this could be split into it's own function. There's a lot going on here and a named function might help identify to new users the functionality in here.

bikespace_frontend/src/components/parking-map/parking-map-page/map-filters/ParkingMapFilter.tsx
```diff
@@ -0,0 +1,295 @@
...
+          Select {enabledFilterPropertiesLookup[filterProperty].description}:
+        </legend>
+        {[...propertyOptions].map(option => (
+          <div key={option}>
+            <input
+              type="checkbox"
+              id={option}
+              name={option}
+              checked={selectedPropertyOptions.has(option)}
+              onChange={e => {
```
MichaelMifsudSweeney: Consider abstracting the onChange to it's own named function.

bikespace_frontend/src/components/parking-map/parking-map-page/map-filters/ParkingMapFilter.tsx
```diff
@@ -0,0 +1,295 @@
...
+
+interface ParkingMapFilterProps {
+  mapRef: RefObject<MapRef>;
+  setFilter: React.Dispatch<React.SetStateAction<FilterSpecification>>;
+  defaultProperty?: string;
+  enabledFilterProperties?: FilterPropertyAttributes[];
+  onlyShowEnabledFilterProperties?: boolean;
+}
+
+export function ParkingMapFilter({
```
MichaelMifsudSweeney, line 117: With the amount of custom logic here. It probably makes sense to make this it's own hook to seperate all the logic from the display.

bikespace_frontend/src/components/parking-map/parking-map-page/map-filters/ParkingMapFilter.tsx
```diff
@@ -0,0 +1,295 @@
...
+          transformPropertyOptions(
+            f.properties?.[filterProperty],
+            'stringValue'
+          ) as string
+      )
+    );
+    const propertyType =
+      enabledFilterPropertiesLookup[filterProperty].type ?? 'string';
+    let newPropertyOptionsSorted: string[];
+    if (propertyType === 'string') {
```
MichaelMifsudSweeney: with all of this logic generation, it may make sense to have named helper functions to abstract away what these are doing it it's more digestable

bikespace_frontend/src/components/parking-map/parking-map-page/parking-map-page.module.scss
tallcoleman: Thanks! Added with https://github.com/bikespace/bikespace/pull/273/commits/3c76270c151a10df21972b5d4aef281f00b8a185

bikespace_frontend/src/components/parking-map/parking-map-page/map-filters/ParkingMapFilter.tsx
tallcoleman: Good suggestion, have split into a separate file with https://github.com/bikespace/bikespace/pull/273/commits/4459d8a540f119ffd9bbda5134b58243a7fd71a8

bikespace_frontend/src/components/parking-map/parking-map-page/map-filters/ParkingMapFilter.tsx
tallcoleman: Should be resolved in https://github.com/bikespace/bikespace/pull/273/commits/16cbbb7c38b84fa90f5e3942ec72d2b78d087847 - have moved part of the logic into a separate sort function

bikespace_frontend/src/components/parking-map/parking-map-page/map-filters/ParkingMapFilter.tsx
tallcoleman: Should be resolved in https://github.com/bikespace/bikespace/pull/273/commits/16cbbb7c38b84fa90f5e3942ec72d2b78d087847 - have moved part of the logic into a separate sort function

bikespace_frontend/src/components/parking-map/parking-map-page/map-filters/ParkingMapFilter.tsx
tallcoleman: Should be resolved in https://github.com/bikespace/bikespace/pull/273/commits/16cbbb7c38b84fa90f5e3942ec72d2b78d087847

tallcoleman:
@MichaelMifsudSweeney I made some changes based on your comments (thank you for the review!) Do you have any interest in doing the final approving review, or should I find someone else to take a look?

MichaelMifsudSweeney (approved):
Nicely done! There's a ton of code added, so would be good to at some point do a large clean, but I don't think there's anything blocking as it is functioning well and there's test coverage. 

If I find some time soon would be fun to take a deeper clean 🧹.

Let me know how testing goes!

tallcoleman:
Thanks for the final review @MichaelMifsudSweeney 

And if you ever want to take a deeper look at some of the existing code... feel free!

PR #276, opened 2025-07-18
Add flask-admin and flask-security

sonalranjit:
Goals:

- Set up admin user accounts (no registration option for general users at this point). Role types should include:
  - Admins (can add/remove users and edit data)
  - Maintainers (can edit data and manage their own account)
- Set up and test admin panels for logging in and managing accounts

Future PRs:

- Set up versioning in the `bikeparking_submissions` table to allow for updates with recorded history
- Add a status table linked to `bikeparking_submissions`

Todos:

- [ ] security review
- [x] update tests and check coverage
- [x] set-up flow: add secrets for seed user


tallcoleman:
@sonalranjit added a script to add a seed superuser using environment secrets and updated the tests so that they are now passing

bikespace_api/bikespace_api/api/forms.py
tallcoleman, line 1: Is this being used or can this file be deleted?

bikespace_api/bikespace_api/api/models.py
```diff
@@ -6,7 +6,8 @@
 from bikespace_api import db
 from datetime import datetime, timezone
 from enum import Enum
-
+from flask_security import RoleMixin, UserMixin
+from werkzeug.security import generate_password_hash, check_password_hash
```
tallcoleman: In my editor, it shows the following errors for the imports:

- sqlalchemy.sql.func - import not used
- RoleMixin and UserMixin should be imported from flask_security.core
- werkzeug.security imports not used

bikespace_api/bikespace_api/api/views.py
```diff
@@ -0,0 +1,107 @@
+from flask import abort, redirect, request, url_for
+from flask_admin import AdminIndexView
+from flask_admin.contrib.sqla import ModelView 
+from flask_security import current_user
+from flask_security.utils import hash_password
+from wtforms import SelectField, PasswordField
+from bikespace_api.api.models import IssueType, ParkingDuration
```
tallcoleman: the bikespace_api.api.models imports as well as the wtforms.SelectField import appear to be un-used

bikespace_api/bikespace_api/api/views.py
tallcoleman, line 1: A lot of the class methods (e.g. `is_accessible` and `_handle_view`) are the same or very similar for these views. Would it make sense to make a custom subclass to use instead of ModelView?

See: https://github.com/pallets-eco/flask-admin/blob/a9f2e0c418f57f97b3f1a59bb8680473418a1c56/examples/auth/main.py#L88

bikespace_api/bikespace_api/templates/security/register_user.html
tallcoleman, line 1: I'm assuming this template is not currently being used but can be for future updates?

bikespace_api/bikespace_api/api/views.py
```diff
@@ -0,0 +1,107 @@
...
+import uuid
+
+class AdminIndexView(AdminIndexView):
```
tallcoleman: When I run `make test-api-terminal` it shows that there's no coverage for this class even though `test_admin_page.py` is calling the admin index page both logged in and not. Do you know why this is the case?

bikespace_api/bikespace_api/api/views.py
```diff
@@ -0,0 +1,115 @@
...
+from flask_admin import AdminIndexView
+from flask_admin.contrib.sqla import ModelView
+from flask_security import current_user  # type: ignore
+from flask_security.utils import hash_password
+from wtforms import SelectField, PasswordField
+from bikespace_api.api.models import IssueType, ParkingDuration
+import uuid
+
+
+class MyAdminIndexView(AdminIndexView):
```
tallcoleman: This function appears to not get any coverage (i.e. is not getting used) even if you call `test_client.get("/admin/")` with a superuser logged in. I think that it needs to be added when `Admin` is configured in the `__init__.py` `create_app` function.

I think if implemented, a regular user could still technically log in, but all the admin pages (including the root "/admin/" would 403 for them. The `test_allowed_pages_for_regular_users` function may need to be updated if that's the intended behaviour.

bikespace_api/bikespace_api/tests/conftest.py
tallcoleman, line 1: FYI I updated the way the fixtures are set up to align with the documentation in https://flask.palletsprojects.com/en/stable/testing/

- having module scope for the test client meant that if a user was logged in in one test, they would also be logged in for others, which led to some confusing consequences for tests in the same module that test login and non-login states, including for users with different permissions
- app context (e.g. required for tests that use direct db queries) can be provisioned as-needed using the flask_app fixture instead of being provided every time the test client is used

bikespace_api/bikespace_api/config.py
```diff
@@ -11,8 +11,28 @@ class BaseConfig:
...
+    SECURITY_REGISTER_URL = "/register/"
+
+    SECURITY_POST_LOGIN_VIEW = "/admin/"
+    SECURITY_POST_LOGOUT_VIEW = "/admin/"
+    SECURITY_POST_REGISTER_VIEW = "/admin/"
+
+    # Flask-Security features
+    SECURITY_REGISTERABLE = False
+    SECURITY_SEND_REGISTER_EMAIL = False
+    SQLALCHEMY_TRACK_MODIFICATIONS = False
```
tallcoleman, line 31: What does this flag do?

bikespace_api/bikespace_api/config.py
```diff
@@ -26,6 +46,7 @@ class TestingConfig(BaseConfig):
 
     TESTING = True
     SQLALCHEMY_DATABASE_URI = os.environ.get("TEST_DATABASE_URI")
+    WTF_CSRF_ENABLED = False
```
tallcoleman, line 49: Added this to the testing config since there are some endpoints that would otherwise require a CSRF token - there are ways to have your tests also get and use a CSRF token, but that's additional complexity we could look at adding in later.

bikespace_api/manage.py
```diff
@@ -78,13 +104,69 @@ def seed_db():
     )
     db.session.commit()
 
+    # add roles for seeded users if they do not yet exist
```
tallcoleman, line 107: Something worth looking into in the future - I think we can do some of this set-up using test fixtures instead; might be cleaner. For now, this is fine though!

tallcoleman (changes requested):
@sonalranjit thanks for all your hard work on this first step towards enabling us to clean up our data and better track the status of issues!

FYI I added some more test coverage to your changes, but it's still slightly below what it was before, so the coveralls action is unhappy. If we're not able to get it back over 98% again, you might have to bypass that check to merge or re-config the coveralls check?

I still want to make sure we do our clean-up non-destructively (i.e. tagging test submissions instead of outright deleting them from the db), so I'll plan to do a quick follow-up PR to make that possible.

bikespace_api/bikespace_api/api/views.py
tallcoleman: see: https://github.com/bikespace/bikespace/pull/276#discussion_r2460192707

bikespace_api/bikespace_api/config.py
sonalranjit: These disable self registering an admin account. We don't want that.

bikespace_api/bikespace_api/api/models.py
sonalranjit: hmm strange, i'll look at it

bikespace_api/bikespace_api/api/views.py
sonalranjit: Yupp, seems to be unused, i'll remove it

bikespace_api/bikespace_api/templates/security/register_user.html
sonalranjit: Yeah not used, im not sure if we want a self registering page

sonalranjit:
Screenshot 2025-10-28 at 5 50 14 PM
Something is up with Flask-Admin and WTForms version, seems like Flask-Admin `1.6.1` isn't compatible with WTForms `3.1.2` When i go to create a new Role it throws an error.

bikespace_api/bikespace_api/api/views.py
sonalranjit: Yeah i removed this view, doesn't seem to do anything

sonalranjit:
This is ready to be reviewed again @tallcoleman, i've deployed it on a test deployment on fly, it seems to be working

tallcoleman (approved):
@sonalranjit I think this is good to merge now. I pushed one fix to one of the tests since the test coverage was indicating that the assert condition wasn't getting called.

sonalranjit:
Awesome thanks @tallcoleman

Issue #223, opened 2024-12-24
Improve data loading for dashboard api request


tallcoleman:
Two improvements that can be made:

## Improve responsiveness for permalinks to a specific issue

Currently, if a user uses a permalink to an issue (e.g. `https://bikespace.ca/dashboard?submission_id=656`), the dashboard has to load all the data before it can display the one issue the user is navigating to.

This could be improved by:

- adding an endpoint to the API to request a specific issue report
- adding logic to the data fetching for the dashboard where if a specific issue is requested in the url param on load, then that issue is requested first, and then the rest of the data is loaded later when it arrives. Would have to check that the number of issues being requested is actually a significant bottleneck (see also: https://github.com/bikespace/bikespace/issues/266 )

## Remove hardcoded API page size

The API call for the dashboard page currently has a hard-coded page size of 5000. We should replace this with the proper logic to return all of the applicable submissions. With about 1,500 submissions currently, this isn't an imminent issue, but should be addressed at some point.

Line 10 in [bikespace_frontend/src/hooks/use-submissions-query.ts](https://github.com/bikespace/bikespace/blob/main/bikespace_frontend/src/hooks/use-submissions-query.ts#L10)

Simplest solution (since the payload of all reports at once is still pretty small):

- GET Request `https://api-dev.bikespace.ca/api/v2/submissions?limit=1`
- Extract the `total_items` value from the response
- GET Request `https://api-dev.bikespace.ca/api/v2/submissions?limit=<<total_items>>`

For either of these changes where the data would be loaded with multiple requests, should make sure that they don't cause any issues with the loading spinner; see: https://github.com/bikespace/bikespace/pull/286#discussion_r2462914815

Yup03:
Hi, I see that this issue is not yet assigned
Can it be assigned to me so i can work on it??

KangarooAlpha:
Hi! I’ve started working on this and here's what I have so far:
- Added a new API call to fetch a single submission when a `submission_id` is present in the dashboard URL.
- Load that submission first so it displays immediately.
- Then fetch the full submissions dataset in the background as before.
- Updated the filtering logic to handle both the single-submission and full-submissions cases cleanly.

While it's faster on the backend, the user experience is still the same, the loading spinner still blocks interaction until the full dataset finishes loading.

I’m considering updating the loading state so that:
- The spinner is dismissed as soon as the single submission loads
- The rest of the UI continues loading submissions in the background.

Before I continue, I wanted to confirm:

[...]

PR #286, opened 2025-09-20
Loading spinner

### LINKED ISSUE #272: Feature request - loading state spinner on parking map
User request: add a loading state spinner on the parking map.

A related observation is that the "feed" tab of the dashboard does not have a loading state indicator - could be fixed with the same PR.

krhobeck:
Added loading spinner to both the dashboard and parking map

Updated sidebar spinner for the dashboard map


bikespace_frontend/src/components/spinner/spinner.module.scss
```diff
@@ -0,0 +1,46 @@
...
+//}
+//
+//@keyframes rotation {
+//  0% {
+//    transform: rotate(0deg);
+//  }
+//  100% {
+//    transform: rotate(360deg);
+//  }
+//}
```
sonalranjit: Maybe we can delete dead code, the changes will be tracked in git if we need to revert it back later

bikespace_frontend/src/components/dashboard/report-summary/report-summary.module.scss
```diff
@@ -21,13 +21,13 @@
...
+//.loading {
+//  min-height: 72px;
+//  padding: 4px 8px;
+//  border-radius: 4px;
+//  background-color: $color-primary-d50p;
+//  color: white;
+//  font-size: 2rem;
+//  font-weight: 600;
+//  // text-align: center;
+//}
```
sonalranjit: Same here, perhaps we can just delete the dead code

sonalranjit:
Thanks @krhobeck ! it's great addition with the spinner. I'll let @tallcoleman add his review, he's got a bit more frontend expertise than me

krhobeck:
No problem, I'll delete that dead code when I can

sonalranjit:
@krhobeck I notice a bit of a lag when zooming in and out, I guessing it's because the markers now re-render once the tiles are loaded. 
screen-recording-1

Compared to what it is currently where the markers stay rendered even when zooming in and out of layers.
screen-recording-2

If possible could we look to maintain the current behavior

krhobeck:
Ah for sure, let me look into fixing this

tallcoleman:
Thanks for doing this @krhobeck ! I'll try to take a look when I have time this week

bikespace_frontend/src/components/dashboard/map/Map.tsx
```diff
@@ -1,23 +1,26 @@
...
 import {useStore} from '@/states/store';
 import {defaultMapCenter} from '@/utils/map-utils';
-
-import 'leaflet/dist/leaflet.css';
-import 'leaflet-defaulticon-compatibility/dist/leaflet-defaulticon-compatibility.css';
-
 import {SubmissionApiPayload} from '@/interfaces/Submission';
 import {useSidebarTab} from '@/states/url-params';
 
+import {Spinner} from '@/components/shared-ui/spinner';
```
tallcoleman, line 13: @krhobeck in #273 I added a components folder named shared-ui and have put the spinner there too. (After updating from main).

Also updated some import stuff:

- For a lot of component imports, we use an index.ts and import from the folder instead of from the component .tsx file. I guess we do this in case there are multiple related things you would want to export from the folder? But it keeps the import paths simpler too.
- Do you know if there are any tools / best practices for ordering imports? Based on reading the code that other frontend BikeSpace devs have written, I've internalized something like the following, but I have no idea how it relates to best practices:
  - framework imports (e.g. React, next)
  - third party dependencies (e.g. plotly, maplibre)
  - internal stuff in other folders (usually @-referenced, sometimes broken out by hooks, components, etc.)
  - internal stuff in the same folder (e.g. related components or ts files; usually referenced via relative path)
  - styles (all together regardless of whether they're in the same folder?)
  - assets (e.g. images, svgs)

bikespace_frontend/src/components/dashboard/map/Map.tsx
```diff
@@ -1,23 +1,26 @@
...
+
 import {MapMarker} from '../map-marker';
 import {LeafletLocateControl} from '../leaflet-locate-control';
 import {MapHandler} from '../map-handler';
 
+import 'leaflet/dist/leaflet.css';
+import 'leaflet-defaulticon-compatibility/dist/leaflet-defaulticon-compatibility.css';
+
 import styles from './map.module.scss';
+import loader from '@/styles/shared/loader.module.scss';
```
tallcoleman: This import appears to be unused

bikespace_frontend/src/components/dashboard/map/Map.tsx
```diff
@@ -32,6 +35,8 @@ function Map({submissions}: MapProps) {
   const markerRefs = useRef<MarkerRefs>({});
 
   const [doneLoading, setDoneLoading] = useState(false);
+  const [tilesLoaded, setTilesLoaded] = useState(false);
+  const isMapLoading = !doneLoading;
```
tallcoleman: it looks like every use of `doneLoading` (including in `MapMarker` passed via prop) is inverting the value, so you could just change the main hook here to `[isMapLoading, setIsMapLoading] = useState(true)`, update the downstream values to un-invert them, and remove this const.

bikespace_frontend/src/components/dashboard/map/Map.tsx
```diff
@@ -44,6 +49,9 @@ function Map({submissions}: MapProps) {
     if (!mapRef.current) return;
     mapRef.current.invalidateSize();
   }, [isSidebarOpen, currentSidebarTab]);
+  useEffect(() => {
```
tallcoleman, line 51: Want to check my understanding of what's happening here - let me know if this sounds right?

- default state is "map is loading"
- if submissions.length changes, flips state to "map is loading"
- only once the final marker is drawn (see function in MapMarker ref on line 86) then state flipped to "map done loading"

I don't know React well enough - is there any chance of a race condition here? Are we sure that the marker rendering is always going to happen at least one cycle after submissions.length changes?

bikespace_frontend/src/components/dashboard/map/Map.tsx
```diff
@@ -59,27 +67,40 @@ function Map({submissions}: MapProps) {
         attribution='&copy; Maps <a href="https://www.thunderforest.com/">Thunderforest</a>, &copy; Data <a href="https://www.openstreetmap.org/copyright">OpenStreetMap contributors</a>'
         url="https://tile.thunderforest.com/atlas/{z}/{x}/{y}.png?apikey=66ccf6226ef54ef38a6b97fe0b0e5d2e"
         maxZoom={20}
+        eventHandlers={{
```
tallcoleman, line 75: This is neat! I should look up whether it's possible to do something similar in MapLibre

bikespace_frontend/src/components/dashboard/map/Map.tsx
```diff
@@ -59,27 +67,40 @@ function Map({submissions}: MapProps) {
...
-                markerRefs.current[submission.id] = m;
-                if (index === submissions.length - 1 && !doneLoading) {
-                  setDoneLoading(true);
-                }
-              }}
-            />
-          );
-        })}
-      </MarkerClusterGroup>
+      {tilesLoaded && ( // Tiles should be loaded before rendering markers
```
tallcoleman: What happens if the markers render before the tiles?

bikespace_frontend/src/components/dashboard/sidebar-content/_SidebarContentInsights.tsx
```diff
@@ -1,13 +1,15 @@
-import React from 'react';
+import React, {Suspense, useCallback, useState} from 'react';
```
tallcoleman: Suspense appears to be un-used

bikespace_frontend/src/components/dashboard/sidebar-content/_SidebarContentInsights.tsx
```diff
@@ -32,12 +34,24 @@ const DataDurationByTodChart = dynamic<ChartProps>(
 );
 
 export function SidebarContentInsights() {
+  const TOTAL = 3;
+  const [readyCount, setReadyCount] = useState(0);
+  const markReady = useCallback(
```
tallcoleman: is useCallback necessary here?

bikespace_frontend/src/components/dashboard/sidebar-content/_SidebarContentInsights.tsx
```diff
@@ -32,12 +34,24 @@ const DataDurationByTodChart = dynamic<ChartProps>(
...
-      <DataIssueFrequencyChart className={styles.chart} />
-      <DataFrequencyByDayChart className={styles.chart} />
-      <DataDurationByTodChart className={styles.chart} />
-    </>
+    <div>
+      {!allReady && (
+        <Spinner overlay label="Loading insights..." style={{zIndex: 3000}} />
+      )}
+      <div style={{visibility: allReady ? 'visible' : 'hidden'}}>
+        <ReportSummary />
```
tallcoleman: Since ReportSummary isn't one of the components that this spinner is waiting for, should it be moved above this spinner section?

bikespace_frontend/src/components/parking-map/parking-map-page/parking-map-page.module.scss
```diff
@@ -15,12 +15,20 @@
 .sideBarContainer {
   overflow-y: scroll;
   margin: 4px;
+  // hide scrollbar (firefox and edge)
```
tallcoleman: My preference would be to add more margin left and right of the sidebar content instead of hiding the scroll bar, since the scroll bar could be considered an accessibility feature for some users. That way if there is a visible scroll bar, it doesn't overlap the content.

bikespace_frontend/src/components/parking-map/parking-map-page/ParkingMapPage.tsx
```diff
@@ -45,7 +46,7 @@ import type {
 
 import 'maplibre-gl/dist/maplibre-gl.css';
 import styles from './parking-map-page.module.scss';
-
+import loader from '@/styles/shared/loader.module.scss';
```
tallcoleman: Import appears to not be used?

bikespace_frontend/src/components/parking-map/parking-map-page/ParkingMapPage.tsx
```diff
@@ -220,6 +223,7 @@ export function ParkingMapPage() {
     if (process.env.NODE_ENV !== 'production') console.log('map loaded');
     addSprite();
     handleMouseHover();
+    console.log('done');
```
tallcoleman: What does this do? There's a similar console.log above for testing, though it doesn't run in production.

bikespace_frontend/src/components/parking-map/parking-map-page/ParkingMapPage.tsx
```diff
@@ -307,6 +311,7 @@ export function ParkingMapPage() {
         }
         onLoad={handleOnLoad}
         onClick={handleLayerClick}
+        onIdle={() => setIsMapLoading(false)} // set to false when map is idle
```
tallcoleman: Might be fine in this case, since this value currently only goes one way, but onIdle will fire every time the map stops doing something, including after pans and zooms that load additional vector tiles. May cause unneccessary re-renders?

bikespace_frontend/src/components/shared-ui/spinner/spinner.module.scss
```diff
@@ -0,0 +1,25 @@
+@use './styles/variables.module.scss' as *;
```
tallcoleman: I think we usually do `@import '@/styles/variables.module.scss';` Is there a reason you used `@use` in this case?

bikespace_frontend/src/styles/shared/loader.module.scss
tallcoleman, line 1: This doesn't seem to be used in any of the files it's being imported by; should it be removed?

bikespace_frontend/src/components/dashboard/map/Map.tsx
tallcoleman: Also, this state check inside the `MapContainer` is what is causing the markers to re-render on every zoom. (The issue that Sonal noted).

Maybe return a separate div with the spinner instead of the `MapContainer` while it's in the initial loading state? That would also ensure the spinner shows before the Map renders - currently, because the spinner is part of `MapContainer`, the spinner doesn't show up until Leaflet starts rendering at least the bare bones of the map.

tallcoleman (changes requested):
Thanks @krhobeck this is really great! Will make sure that users don't think things are broken when they're just loading, especially if they're on mobile data or something slower. Have left some comments for you to take a look at. If you want help with fixing up anything to get it merge-ready, just let us know.

tallcoleman:
> @krhobeck I notice a bit of a lag when zooming in and out, I guessing it's because the markers now re-render once the tiles are loaded.

@krhobeck @sonalranjit have noted what I think is the cause of the issue here: https://github.com/bikespace/bikespace/pull/286/files#r2386171017

bikespace_frontend/src/components/dashboard/sidebar-content/_SidebarContentInsights.tsx
krhobeck: It should be, just added this

bikespace_frontend/src/components/dashboard/sidebar-content/_SidebarContentInsights.tsx
krhobeck: It's actually not if you put a guard on each child, I'll update this

bikespace_frontend/src/components/parking-map/parking-map-page/parking-map-page.module.scss
krhobeck: Sure, good point!

krhobeck: Do you mean padding instead of margin here? Adding a right padding will give the bar its own space on the side instead of merging with the sidebar content

bikespace_frontend/src/components/parking-map/parking-map-page/ParkingMapPage.tsx
krhobeck: Would you prefer if I added an eslint rule to catch these?

bikespace_frontend/src/styles/shared/loader.module.scss
krhobeck: This was a legacy loader design I deprecated, I can remove it

bikespace_frontend/src/components/parking-map/parking-map-page/parking-map-page.module.scss
tallcoleman: Interesting... I didn't realize adding more margin moves the scrollbar too. 

Can you remind me which browser/OS combination you were trying to solve for here? Tried mac and windows x chrome & firefox x system settings on/off for persistent scrollbars and noticed the following:

- Persistent scrollbars seem to take up their own space instead of overlapping (which seems fine?)
- Auto-hiding scrollbars (I think is the default on mac and windows) overlap content while scrolling but hide again once the scroll action stops - there is an overlap potential here, but the current padding on the content boxes is usually good enough. Padding could be bumped up slightly if there's an OS/Browser combo we're aware of that has much thicker auto-hiding scrollbars.

bikespace_frontend/src/components/parking-map/parking-map-page/ParkingMapPage.tsx
tallcoleman: I think it would be just removing the current setting in `bikespace_frontend/.eslintrc.json:10-11`. 

I don't remember why we had that set, but I would guess someone was annoyed by having the linter fail the CI on the PR branch while they're in the middle of building out a feature. I guess people could just comment out an unused import until they use it later if they want to avoid the CI failure?

edit: maybe if it's just set to "warn" it won't fail the CI?

bikespace_frontend/src/components/dashboard/sidebar-content/_SidebarContentInsights.tsx
tallcoleman: Thanks - I don't see any additional commits on the branch, were you planning to push them later?

bikespace_frontend/src/components/parking-map/parking-map-page/parking-map-page.module.scss
krhobeck: So I use brave + Linux mint, perhaps it has something to do with that, I'll push and take a look on my macbook to see if it looks weird. Here's a side by side comparison

image

image

The first image shows a padding-right by 6px, the second image is the padding-right removed with more margin set to 8px. But maybe the second image is the desired look? What are your thoughts?

bikespace_frontend/src/components/parking-map/parking-map-page/ParkingMapPage.tsx
krhobeck: I changed to set it to warn for now

bikespace_frontend/src/components/dashboard/map/Map.tsx
krhobeck: I'm not sure about best practices but the one you listed here sounds good to me, I generally follow the pattern of doing

1. Framework imports
2. Third party dependencies
3. 'Large' imports like large components or state stores or API types
4. 'Small' imports like reusables or any separate functions or types [this could also be small components, I usually put these first]
5. Styles & Assets

(I have a small exception for API types i usually put these closer to the top)

I guess I usually don't order based on folder but based on importance/size. For import sorting we could enforce a policy using eslint for this using a library like https://github.com/lydell/eslint-plugin-simple-import-sort and have regexes for this.

krhobeck:
Hey @tallcoleman I made some new changes, some of the old implementations i've scrapped completely, I decided to just take a simpler approach and keep an initial loading on the page.

bikespace_frontend/src/components/dashboard/map/Map.tsx
tallcoleman: Thanks, I'll look into that plugin. Have started a discussion thread for this here: https://github.com/bikespace/bikespace/discussions/292

bikespace_frontend/src/components/parking-map/parking-map-page/parking-map-page.module.scss
```diff
@@ -15,14 +15,7 @@
...
   margin: 4px;
-  // hide scrollbar (firefox and edge)
-  scrollbar-width: none;
-  -ms-overflow-style: none;
-  // hide scrollbar (chromium)
-  &::-webkit-scrollbar {
-    width: 0;
-    height: 0;
-  }
+  padding-right: 6px;
```
tallcoleman: Can you please set this so it only applies in the desktop layout (i.e. when the viewport width is > $wrapper-full-width)?

bikespace_frontend/src/components/shared-ui/spinner/spinner.module.scss
```diff
@@ -1,4 +1,4 @@
-@use './styles/variables.module.scss' as *;
+@import '@/styles/variables.module.scss';
```
tallcoleman, line 1: I looked this up and realized that you were totally right to prefer `@use` over `@import` - I had no idea that `@import` had been deprecated.

Will open a new issue for converting the usage throughout the codebase since we'll need to think about whether we want to just `import as *` like you've done here or convert to the more default namespaced imports from `@use`.

bikespace_frontend/src/components/parking-map/parking-map-page/ParkingMapPage.tsx
```diff
@@ -201,6 +202,7 @@ export function ParkingMapPage() {
 
   function addSprite() {
     mapRef.current!.addSprite('parking', parkingSpritePath);
+    console.log('loading...');
```
tallcoleman: Can this be removed or does it need to stay in (e.g. for testing purposes?)

bikespace_frontend/src/components/parking-map/parking-map-page/ParkingMapPage.tsx
```diff
@@ -221,6 +222,10 @@ export function ParkingMapPage() {
     if (process.env.NODE_ENV !== 'production') console.log('map loaded');
     addSprite();
     handleMouseHover();
+
+    // after styles load
+    const map: MapLibreMap = mapRef.current!.getMap(); // maplibre-gl map instance
+    map?.once('idle', () => setIsMapLoading(false)); // wait for styles to load, then set loading to false
```
tallcoleman, line 227: I like this. Very elegant!

bikespace_frontend/src/components/dashboard/map/Map.tsx
tallcoleman: Fixed in https://github.com/bikespace/bikespace/pull/286/commits/8857a55323cd9a5d325a69029a34ba72fd4f3f45 (or possibly earlier?)

bikespace_frontend/src/components/dashboard/map/Map.tsx
tallcoleman: Resolved in https://github.com/bikespace/bikespace/pull/286/commits/8857a55323cd9a5d325a69029a34ba72fd4f3f45

bikespace_frontend/src/components/dashboard/sidebar-content/_SidebarContentInsights.tsx
tallcoleman: Resolved in https://github.com/bikespace/bikespace/pull/286/commits/865d3899045fc1db871d2ce2cbf4cc9754934238

bikespace_frontend/src/components/parking-map/parking-map-page/parking-map-page.module.scss
tallcoleman: The additional 6px padding you added in https://github.com/bikespace/bikespace/pull/286/commits/913c1fbcd7a013dc6c4c1125ca0a2be66dd7fb9e seems fine to me, if that resolves it for these OS/browser combinations you've tested. It's slightly asymmetrical, but it feels logical since the right side is the divider between the sidebar and the map.

Have left a separate comment that the added padding should be desktop only.

Thanks!

bikespace_frontend/src/components/parking-map/parking-map-page/ParkingMapPage.tsx
tallcoleman: Resolved in https://github.com/bikespace/bikespace/pull/286/commits/e3bcbde92c830a58614d3492ad0ebc04f2b403be

bikespace_frontend/src/components/dashboard/sidebar-content/_SidebarContent.module.scss
```diff
@@ -58,6 +58,12 @@ h2.cardHeading {
   }
 }
 
+.InnerMultipleSidebarContent {
+  display: flex;
+  flex-direction: column;
+  gap: 6px;
```
tallcoleman, line 65: I would recommend you leave a comment above this gap value to explain the situation it addresses and for which browser/OS combos

tallcoleman (changes requested):
Thanks @krhobeck this is looking really good now.

Have gone through and resolved the old comments I left that are no longer relevant. Please take a look at the following that still need attention, in addition to the ones in this review:

- https://github.com/bikespace/bikespace/pull/286#discussion_r2386146675
- https://github.com/bikespace/bikespace/pull/286#discussion_r2386062881 (no need to answer the race condition question if it's a time suck)

I also pushed a small CSS fix in https://github.com/bikespace/bikespace/pull/286/commits/40a5f77b15f23091754e9f2181732d160890cf90 ; feel free to change/adjust it if you think a different approach would be better.

bikespace_frontend/src/components/shared-ui/spinner/spinner.module.scss
krhobeck: Sure no problem

bikespace_frontend/src/components/dashboard/map/Map.tsx
krhobeck: I won't lie and I say I know react that well either. I think this version of the code could have that risk where the last marker mounts we call `setDoneLoading` to true but then it gets stomped by the `useEffect` and we have a spinner that is stuck on the screen when all the markers are showing. (This is real and could happen if `submissions.length` changed from 10 -> 12 lets say, see below).

I looked into this a bit and `useEffect` runs **after** the commit phase while ref callbacks run for each mount/unmount **during** the commit phase [here](https://stackoverflow.com/questions/75088157/are-refs-guaranteed-to-be-set-when-useeffect-runs). This means if there are no submissions in the array then the loader will literally run forever (which is questionable design).

bikespace_frontend/src/components/dashboard/sidebar-content/_SidebarContent.module.scss
krhobeck: Good catch, this slipped my mind

bikespace_frontend/src/components/parking-map/parking-map-page/ParkingMapPage.tsx
krhobeck: This has been removed in the latest commit

bikespace_frontend/src/components/parking-map/parking-map-page/parking-map-page.module.scss
krhobeck: Updated this in the latest commit, let me know if that was the implementation you were looking for.

bikespace_frontend/src/components/dashboard/map/Map.tsx
tallcoleman: > if there are no submissions in the array then the loader will literally run forever

I guess in this situation we have other problems anyway, so I won't worry about it for this PR, especially since in our current set-up, the data loads once and either succeeds or fails. We do have an issue to look into [changing up the data loading](https://github.com/bikespace/bikespace/issues/223) so I added a note there in case there's an issue with the data loading/updating multiple times while the user is using the page.

tallcoleman (approved):
Thanks again @krhobeck for all your hard work on this! 

I think you should be able to squash merge now - just make sure to clean up the description on the merge commit.

PR #288, opened 2025-09-26
Skip coveralls if they have a service outage

tallcoleman:
Adds `fail-on-error: false` to the deploy-to-fly action. (`fail-on-error: false` is already configured for test-preview) This prevents API deployment from being blocked by a coveralls service outage.

Also linted the workflow files.


sonalranjit (approved):
LGTM

PR #294, opened 2025-11-02
Add flask-smorest for auto-documentation, input validation, and simplified output formatting

tallcoleman:
Adds [flask-smorest](https://flask-smorest.readthedocs.io) (formerly known as flask-rest-api) to simplify and improve a couple key areas of API development:

- Instead of manually editing the openAPI yaml, flask-smorest allows for expected arguments and response types to be defined using [marshmallow](https://marshmallow.readthedocs.io/en/latest/index.html) schemas and pulls other details used for the openAPI docs from function descriptions or blueprint decorator arguments
- The marshmallow schemas are also used to validate the arguments for each route; an automatic error response is sent if the validation fails (e.g. the user submits a string for an argument that is expecting an int)
- The schemas are also used to simplify output formatting, replacing much of the manual packaging of responses that was done previously.

Key changes:

- Replace blueprint and blueprint registration with flask-smorest `Api` and `Blueprint` methods
- Add config flags for versions, paths, and assets used to automatically generate documentation page
- Update view functions in `submissions.py` to add schemas and use `.arguments`, `.response` and `.alt_response` methods provided by flask-smorest Blueprint
- Have also re-worked main `submissions.py` view function to use [flask MethodView](https://flask.palletsprojects.com/en/stable/views/#method-dispatching-and-apis) structure to allow for different schemas to be applied to GET and POST methods
- Removed old routes and assets used for generating documentation page
- Fix for datetime values on flask-admin pages - allows `submitted_datetime` to be nullable and adds a custom wtforms class to prevent default date input from truncating microseconds, resulting in data loss


sonalranjit (approved):
This is awesome @tallcoleman thanks!

PR #296, opened 2025-11-15
Improve data loading for dashboard api request

### LINKED ISSUE #223: Improve data loading for dashboard api request

KangarooAlpha:
Adds `useSingleSubmissionQuerry` function to prioritize loading of permalink for a specific issue.

Adds logic to allow interaction when the desired issue is loaded for a permalink.

Removes hardcoded api page size limit


tallcoleman:
Thanks @KangarooAlpha I've mentioned to the other front-end BikeSpace people that this need a review, so we'll make sure someone looks at it in the coming week.

P.S. assuming you're in Toronto, feel free to come say hi at [Civic Tech Toronto](https://civictech.ca/) if you like

krhobeck (approved):

tallcoleman:
Thanks again @KangarooAlpha - feel free to join the Civic Tech slack (link at the top of https://civictech.ca/) and find us on the `#proj-bikeparking` channel if you want to say hi!

PR #303, opened 2025-11-29
Use docker-compose for local development of BikeSpace

sonalranjit:
Containerized all components of BikeSpace with docker and docker-compose for ease of development.

- `Dockerfile` for `bikespace_frontend`
- `Dockerfile` for nginx to handle reverse proxy between the frontend and api requests
- `docker-compose.yaml` for the orchestrating all the containers/services need to run bikespace locally

The frontend code now connects with the local api instance, serving local data from the database.
Hot reloading when changes are made to the local source code, it is reflected immediately in the docker containers.


tallcoleman:
@sonalranjit thanks for working on this!

I tried it out on a windows computer and wasn't able to get it to work:

- Running `make run` doesn't work since `make` isn't really a windows tool in the same way it is for mac and linux. I think there are some windows ports, but I haven't tried any of them yet.
- I tried running the docker compose command underlying `make run` and it also failed out on the front-end build with "invalid file request node_modules/.bin/acorn". I think maybe somewhere in the node install instructions it's still trying to install a dependency that's platform-specific?

If you don't have a windows computer, I can spend some time later trying to debug the issue.

I'm also curious what the end goal is here. Is it to make one set-up that runs everything for development, but we still use other docker-compose commands for e.g. running tests? Or are we just trying to use docker to solve platform and installation issues? (e.g. you run a docker command and get a prompt where you can run your various development tasks without worrying about versions of python, npm etc.)

thomassth:
At least on local builds, try `docker-compose up --build`

That shouldn't require any `make` commands

Successfully run all instances on mac, haven't tested on windows yet (but it shouldn't matter on WSL?)

sonalranjit:
> @sonalranjit thanks for working on this!
> 
> I tried it out on a windows computer and wasn't able to get it to work:
> 
>     * Running `make run` doesn't work since `make` isn't really a windows tool in the same way it is for mac and linux. I think there are some windows ports, but I haven't tried any of them yet.
> 
>     * I tried running the docker compose command underlying `make run` and it also failed out on the front-end build with "invalid file request node_modules/.bin/acorn". I think maybe somewhere in the node install instructions it's still trying to install a dependency that's platform-specific?
> 
> 
> If you don't have a windows computer, I can spend some time later trying to debug the issue.
> 
> I'm also curious what the end goal is here. Is it to make one set-up that runs everything for development, but we still use other docker-compose commands for e.g. running tests? Or are we just trying to use docker to solve platform and installation issues? (e.g. you run a docker command and get a prompt where you can run your various development tasks without worrying about versions of python, npm etc.)

Yeah the main purpose is make it platform agnostic for development such as python version and npm version. That seems to be where most people have trouble setting up. Whether it be that their npm version doesn't match based on what the project is setup to do. Or the python version as well.

sonalranjit:
> At least on local builds, try `docker-compose up --build`
> 
> That shouldn't require any `make` commands
> 
> Successfully run all instances on mac, haven't tested on windows yet (but it shouldn't matter on WSL?)

We use `make` as a common build tool to consolidate all our different build tools to use one. Since we have npm, pip etc., we utilize make as the build tool to not have to remember all the various commands for each specific part of the project.

PR #317, opened 2026-01-04
Improve dashboard page navigation and issue selection

tallcoleman:
Improves several regressions to navigation functionality on the dashboard page by adding end to end test coverage and changes to improve navigation and issue selection.

## Main Changes:

- Added end to end test coverage for dashboard page focused on key navigation tasks. 
  - Task for loading page with issue already in URL params and a task to select one issue from the feed menu have been added to the end to end testing suite.
  - Tasks for navigating between several submissions within the same map cluster and navigating the dashboard page menu are too flaky and skipped by default. Have kept the auto-skip tests to aid with future development.
  - Added mobile Safari as a tested browser, but limited CI browsers to desktop Chrome and mobile Safari to prevent the action from timing out with the additional end-to-end tests
- Added manage.py logic to allow for load testing the API and frontend at similar levels to production
- Updated logic for loading submissions when page is loaded with issue in URL param and fixed data loading for useSingleSubmissionsQuery
- URL params no longer update each time the submission is changed, since this seemed to trigger a re-render that cancelled the unclustering animation. 
  - Tracking the currently selected submission is now done as part of the normal state store. 
  - Loading the page with a submission in the params is still supported to provide an option to permalink to a specific issue. 
  - The relevant permalink is now accessible from the ID field on the pop-up or feed descriptions.
- Added trigger to scroll to issue in feed tab when new submissions load or mobile breakpoint is passed 
- Changed dashboard navigation logic so that there is a null state when no tabs are selected.

## Other / Incidental Changes:

- Updated nuqs version (from 1.20.0 to 2.8.9) and made corresponding changes to ensure compatibility (e.g. adding a provider wrapper, config changes for testing)
- Add directories to ignore in search in `.vscode/settings.json`
- Improved rendering logic for issue tags in pop-ups and sidebar
- Cleaned up MapMarker and FeedSubmissionItem components to move effects to a single parent component
- Added a loading indicator to feed tab
- Typing changes to state store files
- Added documentation on generating image snapshots for end to end testing that work on multiple OS platforms
- Added guard for old data cases where the same issue type is stored more than once
- Fixed unintended deadlock state in logic to determine if Dashboard map is done loading

## Outstanding issues

- Issue selection from the sidebar feed, especially switching between issues in the same cluster, can still be flaky - may partly be an issue with the react implementation used for Leaflet markerclusters
- Loading insights tab works in manual testing but often fails in automated end to end tests with the insights tab taking too long to load


tallcoleman:
Quick demo of the key changes here: https://cap.tallcolemanprojects.ca/s/m9kwzmvchvpw9x6

sonalranjit (approved):
Awesome work @tallcoleman , thanks for adding tests for the scenarios

PR #325, opened 2026-02-05
Add Level of Traffic Stress map 

sonalranjit:
Adding a new map to display **Level of Traffic Stress (LTS) map**
- new route for the map `lts-map`
- Adding a new map page which loads pmtiles of the LTS data generate for Toronto
- Added Sidebar for map to display selected segments data, filtering, legend and data sources/methodology
- Added service worker to cache requests for pmtiles dataset


bikespace_frontend/src/components/lts-map/lts-map-page/LtsMapPage.tsx
```diff
@@ -0,0 +1,534 @@
...
+      url: 'pmtiles://backup_map/toronto.pmtiles',
+      attribution:
+        '<a href="https://protomaps.com">Protomaps</a> © <a href="https://openstreetmap.org">OpenStreetMap</a>',
+    },
+  },
+  layers: layers('protomaps', namedFlavor('light'), {lang: 'en'}),
+};
+
+const ltsPmtilesUrl =
+  'pmtiles://https://pub-04bdbe3e39bc434bb5fae50c14232971.r2.dev/lts_toronto_filtered_1_4.pmtiles';
```
tallcoleman: @sonalranjit should this be pulled in as an environment variable, similar to what we do with the bike parking map?

See: https://github.com/bikespace/bikespace/blob/0792bf77e9ffc476d63962514eaf11b70317e057/bikespace_frontend/src/components/map-layers/parking/ParkingLayer.tsx#L37

bikespace_frontend/src/components/lts-map/lts-map-page/LtsMapPage.tsx
tallcoleman, line 1: Ideally, the rendering code for the LTS layer would be put in src/components/map-layers but that doesn't have to be done in this PR; can be done when we e.g. add LTS as an optional layer on the parking map.

tallcoleman (approved):
@sonalranjit so excited to have this addition to the site!

Only critical things to take a look at:

- left a comment about whether you want to pull in the pmtiles file via environment variable instead of hardcoding
- take a quick scan through the other changes I pushed and let me know if you want me to revert or adjust any of them (was also able to make some positive tweaks to the parking map as I did my review)

bikespace_frontend/src/components/lts-map/lts-map-page/LtsMapPage.tsx
sonalranjit: Yeah thanks for pointing it out, i've updated the diff to put the pmtiles url in the environment variable.

bikespace_frontend/src/components/lts-map/lts-map-page/LtsMapPage.tsx
sonalranjit: Yeah we have to think about how to add on all the different layers on the map, it could be overwhelming for a user to have a lot stuff on a dashboard. But yeah we should look into that.

sonalranjit:
@tallcoleman I've addressed your review for the pmtiles url. have another look see if it's all good.

sonalranjit:
In draft until i fix the r2 bucket permissions

tallcoleman (approved):
looks great!

PR #342, opened 2026-02-26
Update lts dataset

sonalranjit:
- Set LTS dataset from 02-27-2026, new dataset has updated rules for residential roads with speed limit up to 40km/h to be LTS 1, and alleys are set to LTS 1.


.github/workflows/test-preview.yml
```diff
@@ -150,7 +150,8 @@ jobs:
       MAPTILER_API_KEY: ${{ secrets.MAPTILER_API_KEY }}
       DATA_BICYCLE_PARKING: ${{ secrets.DATA_BICYCLE_PARKING }}
       DATA_BICYCLE_NETWORK: ${{ secrets.DATA_BICYCLE_NETWORK }}
-      DATA_LEVEL_OF_TRAFFIC_STRESS: ${{ secrets.DATA_LEVEL_OF_TRAFFIC_STRESS }}
+      DATA_LEVEL_OF_TRAFFIC_STRESS: https://storage.bikespace.ca/lts_toronto_filtered_02_26_2026.pmtiles
+      DATA_LEVEL_OF_TRAFFIC_STRESS_LAYER_NAME: ${{ secrets.DATA_LEVEL_OF_TRAFFIC_STRESS_LAYER_NAME }}
```
tallcoleman: I feel like layer names are usually relatively stable properties. Are there layers contained within the pmtiles file that we don't want to show? How do the layer names get generated when you update the LTS data?

.github/workflows/test-preview.yml
```diff
@@ -150,7 +150,8 @@ jobs:
       MAPTILER_API_KEY: ${{ secrets.MAPTILER_API_KEY }}
       DATA_BICYCLE_PARKING: ${{ secrets.DATA_BICYCLE_PARKING }}
       DATA_BICYCLE_NETWORK: ${{ secrets.DATA_BICYCLE_NETWORK }}
-      DATA_LEVEL_OF_TRAFFIC_STRESS: ${{ secrets.DATA_LEVEL_OF_TRAFFIC_STRESS }}
+      DATA_LEVEL_OF_TRAFFIC_STRESS: https://storage.bikespace.ca/lts_toronto_filtered_02_26_2026.pmtiles
```
tallcoleman: Should we aim to have a single canonical url for the most recent LTS pull? (i.e. every time we update the data it just writes over the pmtiles file).

Also does the pmtiles file or s3 call have any way of showing the metatadata of when it was last updated? Or would that have to be implemented separately?

bikespace_frontend/public/sw.js
```diff
@@ -1,7 +1,7 @@
 /* global self */
 
 const scriptUrl = new URL(self.location.href);
-const PMTILES_CACHE = 'pmtiles-cache-v2';
+const PMTILES_CACHE = 'pmtiles-cache-02-26-2026';
```
tallcoleman: What's the purpose of updating the cache name here?

tallcoleman (commented):
Thanks @sonalranjit - trying to get a better sense of what the changes do, have left some comments.

.github/workflows/test-preview.yml
sonalranjit: Yeah let me look at it, i think i can set the layername when using tippecanoe. i was leaving that blank so it was generating the layer name on it's own.

.github/workflows/test-preview.yml
sonalranjit: I think it's good to have a history, also it's needed for the cache, if the filename is same and the url is the same, the serviceworker will have the cache from the old file and will serve the old cache from localstorage.

bikespace_frontend/public/sw.js
sonalranjit: Same comment as above, if the file is updated remotely we have to update the cache name otherwise the serviceworker will just serve the cached files.

sonalranjit:
See https://a355fb1d.bikespace-v2.pages.dev/lts-map deployment as the one to review

.github/workflows/test-preview.yml
tallcoleman: Thanks! Would you be able to write up what flags etc. you usually use when you're doing the conversion? We can add it to the readme for bikespace/LTS-OSM or build it into the scripts there.

.github/workflows/test-preview.yml
tallcoleman: A couple things I noticed when digging into this to try to understand it better:

1. On Chrome and Firefox, the service worker does not appear to be intercepting any requests (example from Chrome, current bikespace.ca lts map page):

Screenshot 2026-02-28 at 9 46 41 AM

2. The browsers appear to be caching some of the file requests locally, so we may not need a service worker to ensure that there is local caching and appropriate performance?

3. The response headers for the pmtiles file do include an eTag and other information that would indicate when the local cache should be replaced. We could test updating the file and seeing if browsers successfully clear their local cache and display the updated file version instead. If we wanted to keep historical versions of the pmtiles file, one option would be to have one canonical copy that's updated for production and secondary historical copies in the same bucket.

tallcoleman (approved):
Thanks Sonal!

Let me know when you've added the tippecanoe how to documentation to the LTS repo and if there's anything else I should try to better understand the web worker.

.github/workflows/test-preview.yml
sonalranjit: Thanks for review that! seems like the service worker wasn't caching the partial http requests, i've updated the service worker, it seems to be intercepting the requests now.
And sure we can explore the other configurations, but service workers is what i understand, i don't really want to rush into implementing something new, this PR's scope was just updating the dataset therefore I don't want to pile on something new.
Also I want to keep the implementations as obvious as possible, like not having to keep track of hidden settings, explicitly setting the cache name works on pretty much all browsers. It doesn't require any need to investigate how other browsers handle caching, We're telling the browser, hey there is a new cache name/version use that.
Screenshot 2026-03-01 at 9 01 12 AM

.github/workflows/test-preview.yml
sonalranjit: Yeah i'll add it to the LTS-OSM repo README

sonalranjit:
This should the one to be reviewed https://f30da6ba.bikespace-v2.pages.dev/

PR #349, opened 2026-03-06
Small design fixes for map pages

tallcoleman:
Made a couple small improvements and clean-up changes to the map pages:

- Parking map: set bicycle network to appear underneath road labels for the parking map so that road labels remain visible
- Parking and LTS maps: Ensure that the correct road labels layer name is used depending on whether the base layers are from maptiler or fallback base map style
- Parking map: previously, the "select on map" button in a parking feature description would not center the feature on the map and would become disabled when the feature was selected. Now the button centers the feature on the map and changes to "center on map" when the feature is already selected.
- Submission dashboard: prevent long words (e.g. URLs) in comment text from breaking layout

### DIFF
--- a/.github/workflows/test-preview.yml
+++ b/.github/workflows/test-preview.yml
@@ -142,15 +142,15 @@ jobs:
           retention-days: 30
   publish:
     # Only publish previews for PRs from the same repository
-    if : ${{ github.event.pull_request.head.repo.full_name == github.repository}}
+    if: ${{ github.event.pull_request.head.repo.full_name == github.repository}}
     name: Publish to Cloudflare Pages
     needs: [test-frontend, test-e2e]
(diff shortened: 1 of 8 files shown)


sonalranjit:
Screenshot 2026-03-20 at 4 47 47 PM
For some reason the LTS map isn't rendering on preview link, seems like it's still referencing lts_toronto_filtered_1_4_02_27_2026.pmtiles. did we update the DATA_LEVEL_OF_TRAFFIC_STRESS in the secrets ?

tallcoleman:
@sonalranjit should be fixed now. The preview action was pulling from repository secrets instead of the .env.production config. Fixed now so that it pulls from .env.production

sonalranjit (approved):
LGTM

Issue #388, opened 2026-05-04
Consolidate dual API test paths (Docker vs local venv) and document port allocation


tallcoleman:
## Problem

There are two parallel ways to run API tests:

1. **Docker path**: `make dev-api-test` starts a `compose-test.yaml` stack — self-contained, matches CI, no local Postgres needed
2. **Local venv path**: `make test-api` / `make test-api-terminal` — requires a local Postgres instance, uses a local venv, calls `manage.py` CLI commands for DB setup

Both paths perform identical setup steps (recreate DB → migrate → seed) but are maintained independently. This creates:
- Risk of config drift between paths
- Cognitive overhead ("which command do I run?")
- Duplicated Makefile logic for DB management

Additionally, the port mapping across environments is undocumented:
- Dev: app `8000`, DB `25432` (external)
- Test: app `8001`, DB internal only
- Prodtest: app `8002`

## Proposed Changes

1. **Designate Docker (`make dev-api-test`) as the canonical test path** and update `AGENTS.md` / `README.md` accordingly. The local venv path is still useful for CI (where Docker-in-Docker has overhead), so label it clearly as the CI/lightweight alternative rather than removing it.

2. **Add comments to the Makefile** explaining the port allocation rationale (e.g., why 8000/8001/8002, why the dev DB is on 25432 instead of 5432).

3. **Extract shared DB setup logic** (recreate → migrate → seed) into a reusable script or Makefile macro so it doesn't need to be duplicated between the two paths.

## Notes

- The local venv path is used in CI (`.github/workflows/`) — check before removing or gating it.
- This is primarily a documentation and maintenance issue, not a functionality one.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #385, opened 2026-05-04
Migrate from legacy SQLAlchemy .query() API and marshmallow _declared_fields to current patterns


tallcoleman:
## Problem

The codebase runs SQLAlchemy 2.0.37, which deprecates the `Model.query` property in favour of `session.execute(select(...))`. There are ~7 uses of `.query` in `submissions_routes.py` and tests:

- `Submission.query.order_by(...).paginate()` (routes, 2×)
- `Submission.query.all()` (routes)
- `Submission.query.filter_by(id=...).first()` (routes + test)
- `SubmissionVersion.query.filter_by(...).order_by(...)` (routes)
- `Submission.query.filter_by(comments=...).first()` (test)

Additionally, `_declared_fields` (a private marshmallow API) is accessed 3× in `submissions_routes.py` to dynamically build CSV and changeset schemas at lines ~88, ~238, and ~298. This is not part of marshmallow's public API contract.

### SQLAlchemy `.query` → `session.execute(select(...))`

```python
# Before
Submission.query.filter_by(id=submission_id).first()

# After
db.session.execute(select(Submission).filter_by(id=submission_id)).scalar_one_or_none()
```

For paginated queries:
```python
# Before
Submission.query.order_by(...).paginate(...)

# After
db.paginate(select(Submission).order_by(...), ...)
```

### `_declared_fields` → public API

Replace `Schema._declared_fields` introspection with either:
- `schema_instance.fields` (on an instantiated schema)
- Explicit field definitions in the schema class body (preferred for discoverability)

The `SubmissionCSVSchema` and changeset schema are dynamically created via `ma.Schema.from_dict(SubmissionSchemaWithVersion._declared_fields | {...})` — these could be restructured as explicit schema subclasses.

## Notes

- The test `test_get_submissions_geojson_empty_db` patches `Submission.query` directly and must be updated to mock `db.session.execute` instead.
- All changes are mechanical but should be done in one focused PR with the full test suite as the verification gate.
- The `.query` API still works in SQLAlchemy 2.0 (it is deprecated, not removed), so this is low urgency but worth cleaning up before a future SQLAlchemy upgrade.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #389, opened 2026-05-04
Reorganize API into modules, add 100% test coverage, fix test isolation

### LINKED ISSUE #386: Add database isolation to pytest fixtures to prevent order-dependent test failures
## Problem

Tests that write to the database (`test_post_submissions`, `test_get_submission_history`, `test_rollback_change`, and most admin tests) share a single live test database with no per-test cleanup. `conftest.py` creates model objects in fixtures but never sets up or tears down tables, and never rolls back transactions between tests.

This means:
- Tests can pass or fail depending on run order
- Leftover rows from one test can affect assertions in another
- The test suite implicitly depends on `manage.py seed_db` having run exactly once before tests start

## Proposed Changes

Add a database isolation fixture using SQLAlchemy's nested transaction (SAVEPOINT) pattern, which wraps each test in a rolled-back transaction:

```python
@pytest.fixture()
def db_session(flask_app):
    with flask_app.app_context():
        connection = db.engine.connect()
        transaction = connection.begin()
        # bind session to this connection so test DB writes use the same transaction
        db.session.bind = connection
        yield db.session
        transaction.rollback()
        connection.close()
```

Alternatively (simpler but slower): a `function`-scoped fixture that calls `db.drop_all()` + `db.create_all()` + seeds minimal data before each DB-using test.

Tests that need specific pre-existing data should create it themselves via fixtures rather than relying on `seed_db` state.

## Notes

- `sqlalchemy_continuum` (used for `SubmissionVersion`) adds complexity here — its version tables must also be cleaned up correctly. The nested transaction approach may interact poorly with continuum's transaction hooks; this will need testing.
- The simpler drop/recreate approach is more reliable in the face of continuum, at the cost of ~1s per DB test.
- Either approach removes the implicit dependency on `seed_db` having run, making the test suite more self-contained.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### LINKED ISSUE #387: Separate dev database seeding from test database seeding in manage.py
## Problem

`manage.py seed_db` does two incompatible jobs:

1. Populates a dev database with realistic data, including **1,500 Faker-generated submissions** for load testing
2. Is called on every test container startup via `boot_local.sh` (`db upgrade → add-seed-user → seed-db → run`)

This makes the test container slow to initialise and couples test infrastructure to dev data volume. The 1,500-record Faker load is unnecessary for running the test suite.

## Proposed Changes

Split `seed_db` into two commands:

- `seed_dev_db` — current behaviour (4 manual submissions + users + 1,500 Faker submissions)
- `seed_test_db` — minimal fixture data only (4 submissions, 2 users: one admin, one non-admin)

Update `boot_local.sh` to call `seed_test_db` instead of `seed_db` so the Docker test environment starts faster.

Keep `seed_dev_db` (or rename the existing `seed_db`) as the command used by `make dev-api`.

Update the Makefile targets accordingly.

## Notes

- The admin tests depend on specific seeded users (`admin@example.com` / `admin` and `notanadmin@example.com` / `notanadmin`) — `seed_test_db` must include these.
- This is a prerequisite for or complement to adding proper database isolation to pytest fixtures (see related issue #386), since minimal seeded data reduces the cost of per-test teardown/recreation.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

tallcoleman:
## Summary

- Reorganizes the API from a flat structure into `submissions/` and `admin/` modules (models, routes, views, tests, templates each in their own module)
- Upgrades all models to SQLAlchemy 2.0 mapped-column style with explicit `back_populates`
- Achieves 100% statement and branch coverage for the API
- Fixes several bugs: `datetime.now()` evaluated at import time, `.env` loading in Docker Compose files, missing `return` in `test_post_submissions`, deprecated `db.backref()` usage, unnamed FK constraint causing `downgrade-db` to fail

Closes #386, closes #387.

Issues #385 and #388 were opened during this work to track remaining follow-up items that can be resolved in separate PRs.

## Changes by area

**Module reorganization**
- `bikespace_api/submissions/` — submission models, routes, admin views, and tests
- `bikespace_api/admin/` — user/role models, admin views, templates, and tests
- `bikespace_api/seed.py` — shared seeding logic (single source of truth for dev and test data)
- `bikespace_api/conftest.py` — moved to package root for auto-discovery; expanded with new fixtures

**Test infrastructure (closes #386, #387)**
- `_setup_database` session fixture: drops, recreates, and seeds the test DB once per session — no more manual `manage.py seed_db` before running tests; uses drop-and-recreate (not create-if-not-exists) so schema-level state can't persist across sessions
- `clean_db` fixture: TRUNCATEs all tables and re-seeds before each DB-writing test, eliminating order-dependent failures
- `seed_dev_db` / `seed_test_db` split in `manage.py`: test containers no longer generate 1,500 Faker submissions on startup
- `logged_in_admin_client` and `new_submission` fixtures remove repeated boilerplate

**Migration fix (incidental)**
- Named the FK constraint in the initial migration (`migrations/versions/8a772e570759_.py`) so `make downgrade-db` no longer fails — discovered while verifying the test plan round-trip

**Documentation**
- Added `AGENTS.md` and `CLAUDE.md` with architecture overview, common commands, and deployment info
- Updated `bikespace_api/README.md` to reflect new module paths and preferred make targets
- Added prerequisite comments in `Makefile` for `upgrade-db` and `downgrade-db` targets

## Test plan

- [x] `make test-api-terminal` passes with 100% coverage
- [x] `make dev-api` starts successfully and API is reachable at `localhost:8000`
- [x] Admin UI accessible and functional (login, create/update user, view submissions)
- [x] `make upgrade-db` / `make downgrade-db` round-trips cleanly with the new migration


sonalranjit (approved):
LGTM! Thanks for reorganizing and increasing the test coverage for the API @tallcoleman

Issue #312, opened 2025-12-10
Allow users to report issue with bike parking from parking map


tallcoleman:
## Goal

Allow users to report a bicycle parking issue from the [bicycle parking map page](https://bikespace.ca/parking-map). There are likely two interactions that need to be implemented:

1. The user selects an existing bicycle feature on the map; in this case, the location for the submission form should be pre-filled. This could be a flow where the user selects a feature and then selects a "report issue" button on the feature card, or the user selects a feature on the map as part of the reporting flow, or both.

2. The user selects a location on the bicycle parking map where there is no bicycle parking feature (e.g. to report bike parking missing or to report that there is bike parking at that location that needs to be added to the data).

The simplest implementation may be as follows, though other options may be worth exploring with the design team:

* Add UI to bicycle parking map to report from a selected feature or to select a location to report an issue
* Navigate the user to the submission page, using a URL param to pre-fill the location field (update the submission form to support pre-filled answers via param)
* After a successful submission, provide an option to go back to the bicycle parking map

It may also be possible to implement the submission form within the bicycle parking map page with a modal window that re-uses the submission form code. If the submission form is implemented as a modal on the bike parking page, it would be preferable to keep the UI consistent with the parking map page instead of using the design and styling from the current submissions page (since that styling is due to be reviewed and made consistent with the rest of the web app).

## Background

Currently to report a bicycle parking issue, users have to use the [submission form](https://bikespace.ca/submission). If a user is on the bicycle parking page, this would mean navigating back or using the "Report Bike Parking Issue" option in the page menu. From there, they would have to re-find the location of the bicycle parking feature that they may have previously selected in the bicycle parking map.

## Potential steps to get started

- Familiarize yourself with the [BikeSpace frontend codebase](https://github.com/bikespace/bikespace/blob/main/bikespace_frontend/README.md)
- Familiarize yourself with [react-map-gl](https://visgl.github.io/react-map-gl/) and Maplibre (map library used on the bicycle parking map)
- Review the existing code for the [submission form](https://github.com/bikespace/bikespace/tree/main/bikespace_frontend/src/components/submission) and the [location input question](https://github.com/bikespace/bikespace/tree/main/bikespace_frontend/src/components/submission/location)

## Dealing with clustered bike parking features

[...]

PR #390, opened 2026-05-04
Implement report issue from parking map

lifers:
In response to #312 .

- [x] Report issue from selected bike parking group
- [x] Report issue from selected location on parking map


lifers:
So far, we have a working prototype of the map autofill in submission form by passing the coordinates in the URL. The goal no. 2 where user can select an empty spot and report an issue requires a new UI item which currently lives in ParkingMapPage.tsx, not sure if we need to factor it out into a component.

tallcoleman:
Question for review - adding a link to bring the user back to the parking map, and how much effort to put into restoring the previous parking map state (e.g. zoom level, map center, items selected)

bikespace_frontend/src/components/parking-map/parking-map-page/ParkingMapPage.tsx
```diff
@@ -273,8 +311,34 @@ export function ParkingMapPage() {
                 handleHover={handleFeatureHover}
                 handleUnHover={handleFeatureUnHover}
                 centerFeatureOnMap={zoomAndFlyToSingleFeature}
+                onReportIssue={handleReportIssue}
               />
             ))}
+            {emptyLocationSelected && parkingGroupSelected.length === 0 && (
+              <div className={styles.ContentCard}>
+                <h3>Selected location has no parking.</h3>
```
tallcoleman: Maybe "Selected location has no known parking."? Or "No known bicycle parking here."?

In a case where someone is reporting that there is bike parking not yet on our map it would be strange to confidently tell the user that we know there's no parking at that location.

bikespace_frontend/src/components/parking-map/parking-map-page/ParkingMapPage.tsx
```diff
@@ -114,6 +117,24 @@ export function ParkingMapPage() {
...
+    lat: number;
+    lon: number;
+  } | null>(null);
+  useEffect(() => {
+    if (!emptyLocationSelected || !mapRef.current) return;
+    const coords = {
+      lon: emptyLocationSelected.lon,
+      lat: emptyLocationSelected.lat,
+    };
+    const marker = new Marker({color: '#a000cc'})
```
tallcoleman: I had an idea for this marker - maybe something like this? I can add it to the sprite sheet if you want.

image

Or if you want to keep the plain marker, I would colour it #ED145B to keep it in the brand palette (see: https://drive.google.com/file/d/1ksqMJAFb-xGw7rRKb7SFBjKEvwJL5D0W/view?usp=drivesdk )

bikespace_frontend/src/components/parking-map/parking-map-page/ParkingMapPage.tsx
```diff
@@ -241,6 +273,12 @@ export function ParkingMapPage() {
     map?.once('idle', () => setIsMapLoading(false)); // wait for styles to load, then set loading to false
   }
 
+  const openSubmission = useSubmissionPrefill();
+  function handleReportIssue(feature: Feature) {
+    const [lon, lat] = getCentroid(feature);
```
tallcoleman, line 278: This is good; not all of the OpenStreetMap features are points, so this should make it more robust (I think the script does simplify to centroids currently but that may change in the future)

bikespace_frontend/src/components/parking-map/parking-map-page/ParkingMapPage.tsx
```diff
@@ -273,8 +311,34 @@ export function ParkingMapPage() {
...
                 handleUnHover={handleFeatureUnHover}
                 centerFeatureOnMap={zoomAndFlyToSingleFeature}
+                onReportIssue={handleReportIssue}
               />
             ))}
+            {emptyLocationSelected && parkingGroupSelected.length === 0 && (
+              <div className={styles.ContentCard}>
+                <h3>Selected location has no parking.</h3>
+                <div className={sidebarDescStyles.featureDescriptionControls}>
+                  <SidebarButton
```
tallcoleman: The parking feature action buttons are styled to be less prominent because they are intended to be a secondary action (the primary action being reading the content and selecting different parking features) But in this case, the issue reporting option is the primary action - is there a way to have the button use the default sizing and border colour instead of the washed out style used for the parking features?

tallcoleman (changes requested):
Thanks for working on this!

Two additional things to consider:

- I think it would be helpful to have a post-submission button that takes the user back to the parking map (if the submission form Lon Lat are pre-filled only). I think a reasonable scope would be to return the user to the centre of the map that they were looking at before at zoom level 18 (same as zoomAndFlyTo). Restoring the users past selection would be harder so I think we can leave that aside for now. 

- Think about what kind of tests would be useful for the features that have been added? Maybe one end to end test that loads the submission form with a URL parameter, and then the rest can be unit tests with Jest. we generally don't try for 100% coverage on front end tests, just things that reflect the key user interactions.

bikespace_frontend/src/components/parking-map/parking-map-page/ParkingMapPage.tsx
lifers: I think "No known bicycle parking here." makes more sense, just amended it.

bikespace_frontend/src/components/parking-map/parking-map-page/ParkingMapPage.tsx
lifers: Removed the featureDescriptionControls styling, now using default SidebarButton style.

bikespace_frontend/src/components/parking-map/parking-map-page/ParkingMapPage.tsx
lifers: It's a small change in the typescript if we already have the sprite, for now I keep the default marker and change the color o follow the brand.

tallcoleman (approved):
Great work, thank you!

I think you should be able to merge the commit now - please squash merge and make sure to edit the merge commit comment.

Issue #382, opened 2026-04-29
Add comment and status system for issues to help track when/if they have been resolved


tallcoleman:
## Background

Currently, users can report issues with bike parking (e.g. bike parking is damaged, there is no bike parking near a certain destination) using the `submissions/` page. These submissions are then viewable on the `dashboard/` page.

BikeSpace volunteers have been following up on many of the issues reported, though this is currently tracked in a spreadsheet. Users outside the volunteer team do not have any visibility into the status of the issues they have reported, actions taken to resolve them, or whether the issue has been resolved. This also means that users cannot contribute their own observations of the issue status and share if they have also taken any steps to try to resolve the issue (e.g. reporting through 311, communicating with their local city councillor).

The goal of this series of tasks is to:

- Add the ability to update the status of an submission (authentication required)
- Display statuses on the `dashboard/` page and use them to help filter the default set of issues shown to be more useful
- Add the ability to leave comments about a submission (authentication optional)
- Display these comments about a submission when the user selects them on the `dashboard/` page
- Since the status updates above require setting up authentication for certain API endpoints, it may also be relatively easy to allow users to optionally authenticate when making a submission (which would enable future updates to allow them to track the status of their submissions specifically)

Relevant user comments from past interviews / usability testing:

- "Having a status of the issue whether it has resolved or not would be great"
- "displaying resolved issues could motivate users."

## Implementation details

Past notes:

- API changes: see [Proposal for Discussion_ API and Database Enhancements](https://docs.google.com/document/d/17ncE0AV2XDxZKqDKe00Aqax9qWjezaGY2t-vXnJEwL0/edit?usp=sharing)
- Design ideas: see https://civictechto.slack.com/archives/C61CZLA5V/p1759109092664879 and https://civictechto.slack.com/archives/C61CZLA5V/p1756956574752259

## Proposed Task List

- [x] Add user accounts and admin panel - via [add flask-admin and flask-security](https://github.com/bikespace/bikespace/pull/276)
- [x] [Add flask-smorest for auto-documentation, input validation, and simplified output formatting](https://github.com/bikespace/bikespace/pull/294) - simplifies future API work
- [ ] Add data models for statuses and comments (some work done on the `api-add-statuses-comments` branch, but needs to be revisited)
- [ ] Add authentication to certain API endpoints and enable a login flow via the front-end (start with optional auth for an existing endpoint?)
- [ ] Finalize list of statuses to be used for submission management
[...]

PR #396, opened 2026-05-09
Add authentication to API endpoints

tallcoleman:
Enabling step for https://github.com/bikespace/bikespace/issues/382 since it will allow for certain endpoints to update the status of a submission to only be available to authenticated users.

Deployment plan (see comments chain) should be reviewed as well as code changes.

## Key Changes

Authorization and route changes:
- Token authentication is now enabled for the API and documented in the OpenAPI docs. Tokens are set to have a 1h life.
- Token authentication is optional for the submissions POST route - if authenticated, the user is recorded on the submission
- Two new routes, both of which require authentication and superuser or editor roles (see notes on roles below): PATCH and DELETE for a single submission
- PATCH route still sanitizes the comment field, same as POST
- Deleted submissions are still viewable through the version history endpoint (existing functionality, but worth noting in this context)
- API configured to exclude token authentication from CSRF checks (does not apply since no cookie is used), but keep CSRF protections for other (cookie-based) auth methods

User model changes:
- Updated `User` model and flask-security configuration to enable usernames and replaced email (potentially PII) as the default `User` string representation with username.
- First Name and Last Name fields for users are now optional

User role changes:
- User roles are now documented in `bikespace_api/bikespace_api/admin/roles.py` with an enum that can be used throughout the codebase to ensure that the role strings remain consistent.
- Added "editor" role that can manage all submissions (and in the future comments and statuses) but not manage roles and users. This will enable additional volunteers to help with data management without needing to provide full admin access.

Submission model and route changes:
- issues field for submissions is now required (matches existing frontend validation)
- validation added on API to prevent list of issues longer than number of issue types
- validation added on API to prevent submitting a comment greater than 5000 characters (added as part of security review)
- optional user relationship added as noted above

## Other changes

- Added test coverage for all new additions/changes
- Added database migrations for changes to user and submission models
- Enabling usernames requires adding `bleach` dependency
- Reorganized and added comments to config to make it clearer which config options relate to which plugins and functionality
- Moved `logged_in_admin_client fixture to root conftest and added fixtures to get auth token for headers and provide a dummy submission with user specified
- Disabled deprecated flask-security anonymous user object and updated user-checking code to add guards for when user is None
- Various style changes including using `HTTPStatus` for different HTTP codes, simplifying code for returning and handling JSON from endpoints, using `Optional` instead of `| None`, using `pytest.mark` instead of just importing `mark`, organizing long test files with classes, using `isinstance` for type assertions, and ensuring all status words in JSON responses are lowercase
- Fix bug with Meta only property being used for subclassed marshmallow models - this does not work with subclassing, only when the model is instantiated. The exclude property should be used for subclassing instead.
- Tweak pragma so that local api test coverage numbers agree with CI

Final tasks:

- [x] Security review (sent separately)
- [x] Make backup and monitoring plan for deploy


tallcoleman:
# Notes on testing requests for using authentication:

(Written as httpie cli commands)

## Authenticate and retrieve token

Request:

POST request to `localhost:8000/admin/login/?include_auth_token` with body `{'email': 'admin@example.com', 'password':'admin'}`

```sh
http POST :8000/admin/login/ \
include_auth_token== \
> email=admin@example.com \
> password=admin
```

Response:

```json
{
    "meta": {
        "code": 200
    },
    "response": {
        "csrf_token": "IjcxNGI0NDE4NDAwYjE2MWJmZTZkMWVhZDM2ZjI0YjJmMjI2Y2I3YzUi.af8xCw.a1_iUPc08rkrAFd4_xfqL74FZPw",
        "user": {
            "authentication_token": "eyJ2ZXIiOiI1IiwidWlkIjoiYzE1MTRjZjdhZTRmNGQ2ZWJiNTM1ZDM3NmI4ODdlNGUiLCJmc19wYWEiOjE3NzgzMzE5MTUuMzc1MzIzMywiZXhwIjowfQ.af8xCw.QEQTHKW39hmLNn9GPuu7Mb4lOFA"
        }
    }
}
```

## Make a request with token authentication

Using `Authentication-Token` header

```sh
http :8000/api/v2/submissions \
> Authentication-Token:eyJ2ZXIiOiI1IiwidWlkIjoiYzE1MTRjZjdhZTRmNGQ2ZWJiNTM1ZDM3NmI4ODdlNGUiLCJmc19wYWEiOjE3NzgzMzE5MTUuMzc1MzIzMywiZXhwIjowfQ.af8xCw.QEQTHKW39hmLNn9GPuu7Mb4lOFA
```

Usual response returned

## Request from auth required endpoint without authenticating

Request: `$ http :8000/api/v2/submissions`

Response is a 302 FOUND:

```http
HTTP/1.1 302 FOUND
Access-Control-Allow-Origin: *
Connection: close
Content-Length: 263
Content-Type: text/html; charset=utf-8
Date: Sat, 09 May 2026 13:06:18 GMT
Location: /admin/login/?next=/api/v2/submissions
Server: Werkzeug/3.1.6 Python/3.12.13
Set-Cookie: session=eyJfZmxhc2hlcyI6W3siIHQiOlsiZXJyb3IiLCJZb3UgbXVzdCBzaWduIGluIHRvIHZpZXcgdGhpcyByZXNvdXJjZS4iXX1dfQ.af8xSg.4uk1H4SsIaq2y_pkvMUF3mhabMM; HttpOnly; Path=/
Vary: Cookie

<!doctype html>
<html lang=en>
<title>Redirecting...</title>
<h1>Redirecting...</h1>
<p>You should be redirected automatically to the target URL: <a href="/admin/login/?next=/api/v2/submissions">/admin/login/?next=/api/v2/submissions</a>. If not, click the link.
```

## Next Steps

- Review documentation and config to require CSRF token with token authentication
- Is 302 FOUND the right response for no auth request? May make sense for admin app but not for API endpoints

tallcoleman:
## Deployment Plan

### Snapshot Backup

Creating the snapshot:

```bash
# record volume id
fly volumes list --app <postgres-app-name>

# create a snapshot for the postgres volume
fly volumes snapshots create <volume-id>

# confirm snapshot is created and record snapshot id
fly volumes snapshots list <volume-id>
```

Restoring (if needed):

```bash
# use scale command to launch new machine using snapshot
fly scale count 0 --app <postgres-app-name>
fly scale count 1 --with-new-volumes --from-snapshot <snapshot-id> --app <postgres-app-name>

# verify restore
# confirm machine running
fly status --app <postgres-app-name>
# check for migration/startup errors
fly logs --app <postgres-app-name>
# test api manually to confirm service is live

# remove old volume once restore confirmed successful
fly volumes delete <old-volume-id>
```

### Password Recovery

As part of the security review, the `BIKESPACE_SECURITY_PASSWORD_SALT` secret in fly was updated. If this prevents users from logging in:

```bash
# start console in app
fly ssh console --app <api-app-name>

# in console use manage.py to reset to temporary new password
python manage.py users change_password <email> --password <temppassword>

# inform user of password change and have them reset it themselves manually once logged in with their temporary password
```

tallcoleman:
Forgot to add a note on the expected impacts of the db migrations:

- Convert bikeparking_submissions.issues to required (not nullable), backfill with ["other"]: Affects 1373, 1372, 1301, 26, 27, 16. These have all been identified as low-quality data in the past (likely test entries, but do not have "test" in comments), so impact is low.
- Add username to User model: affects existing admin users, usernames will be backfilled using First Name, Last Name, and row ID. No issues expected with backfill logic based on existing entries.
- Add optional user field to submissions: no records affected

sonalranjit (approved):
Awesome work @tallcoleman !

tallcoleman:
Post-merge note: `["other"]` was not backfilled for issues 1373, 1372, 1301, 26, 27, and 16. I suspect that the db value for `issues` in these reports is actually `'{}'` (empty array) rather than Null. There is not much value in backfilling `["other"]` for these reports unless it would be worthwhile to remove the guard for the empty array case on the front-end, so I am leaving it as-is for now. The backfill for usernames appeared to work as intended.

Issue #253, opened 2025-05-11
Add Issue Type filter to Filters Tab on Dashboard & Convert check-boxes to chips


portolosalexandra:
Please see the following suggested design. 

The motivation for this is to 1) Modernize the look and feel of the filters tab (Using multi-select chips instead of checkboxes) and 2) To add a filter for the 'Issue Type'. 

Once the issue type filter is added, it probably would also make sense to remove the filter from the data/insights chart?

Suggested Design:

Figma file: https://www.figma.com/design/GN8BdmaWZeWRs0qMW5eccVvk/BikeSpace-dashboard-mockup?node-id=785-2287&p=f&t=6vEuNwLF0rCeXPf2-0

Web:

<img width="581" alt="Image" src="https://github.com/user-attachments/assets/20cee1f7-d42a-4efb-abb9-9cbaeea2dd5f" />

Mobile:

<img width="243" alt="Image" src="https://github.com/user-attachments/assets/c6ccb8b3-4d85-41dc-96b2-203dea2ea66b" />

portolosalexandra:
Hey @tallcoleman and @sonalranjit - thanks for the feedback on this design. I didn't know which ticket to add it to so I created a new one. Feel free to edit / add labels / assign as needed (as I was a bit unsure myself on how best to setup the ticket). 

Thanks!

tallcoleman:
Thanks @portolosalexandra !

PR #409, opened 2026-05-31
Added issue filter

KarishmaChanglani:
For Issue #253, Updates the page to have more intuitive issues filter. For the drop down I didn't see any list of options so I basically added
Match Any, 
Match All, 
Exclude

I think that should suffice?

### DIFF
--- a/bikespace_frontend/e2etests/filter-issue-type.spec.ts
+++ b/bikespace_frontend/e2etests/filter-issue-type.spec.ts
@@ -18,6 +18,22 @@ async function getReportCount(page: Page): Promise<number> {
   return Number(text.replace(/,/g, ''));
 }
 
+// The filtered count updates via an async effect reacting to filter state,
+// so it can lag a render behind the click that triggered it. Poll until two
+// consecutive reads agree before treating the value as settled.
+async function getStableReportCount(page: Page): Promise<number> {
+  let previous: number | undefined;
+  await expect
+    .poll(async () => {
+      const current = await getReportCount(page);
+      const stable = current === previous;
+      previous = current;
+      return stable;
+    })
+    .toBe(true);
+  return getReportCount(page);
+}
+
 test.describe('Issue type filter', () => {
   test('filter mode select is disabled until an issue chip is selected', async ({
     page,
@@ -71,27 +87,35 @@ test.describe('Issue type filter', () => {
   test('filter mode changes which submissions are shown', async ({page}) => {
     const filterModeSelect = page.getByLabel('Filter mode');
 
-    const totalCount = await getReportCount(page);
+    const totalCount = await getStableReportCount(page);
 
-    await page
-      .getByRole('button', {name: 'Abandoned bicycle', exact: true})
-      .click();
-    await page
-      .getByRole('button', {name: 'Parking damaged', exact: true})
-      .click();
+    const abandonedChip = page.getByRole('button', {
+      name: 'Abandoned bicycle',
+      exact: true,
+    });
+    const damagedChip = page.getByRole('button', {
+      name: 'Parking damaged',
+      exact: true,
+    });
+    await abandonedChip.click();
+    await expect(abandonedChip).toHaveAttribute('aria-pressed', 'true');
+    await damagedChip.click();
+    await expect(damagedChip).toHaveAttribute('aria-pressed', 'true');
 
     // default mode is "any": matches submissions with at least one selected issue
     await expect(filterModeSelect).toHaveValue('any');
-    const anyCount = await getReportCount(page);
+    const anyCount = await getStableReportCount(page);
     expect(anyCount).toBeLessThanOrEqual(totalCount);
 
     // "all" is a strict subset of "any": matches submissions with every selected issue
     await filterModeSelect.selectOption('all');
-    await expect.poll(() => getReportCount(page)).toBeLessThanOrEqual(anyCount);
+    const allCount = await getStableReportCount(page);
+    expect(allCount).toBeLessThanOrEqual(anyCount);
 
     // "exclude" matches submissions with none of the selected issues,
     // so it is the complement of "any" (relative to the unfiltered total)
     await filterModeSelect.selectOption('exclude');
-    await expect.poll(() => getReportCount(page)).toBe(totalCount - anyCount);
+    const excludeCount = await getStableReportCount(page);
+    expect(excludeCount).toBe(totalCount - anyCount);
   });
 });
--- a/bikespace_frontend/e2etests/filter-issue-type.spec.ts
+++ b/bikespace_frontend/e2etests/filter-issue-type.spec.ts
@@ -1,0 +1,97 @@
+import {test, expect, Page} from '@playwright/test';
+
+test.use({
+  viewport: {height: 800, width: 1200},
+});
+
+test.beforeEach(async ({context, page}) => {
+  // test isolation: block all network requests except for localhost
+  await context.route(/https?:\/\/(?!localhost).+/, route => route.abort());
+
+  await page.goto('/dashboard?tab=filters');
+  await page.waitForSelector('div.leaflet-container');
+  await expect(page.getByTestId('report-summary-count')).toBeVisible();
+});
+
+async function getReportCount(page: Page): Promise<number> {
+  const text = await page.getByTestId('report-summary-count').innerText();
+  return Number(text.replace(/,/g, ''));
+}
+
+test.describe('Issue type filter', () => {
+  test('filter mode select is disabled until an issue chip is selected', async ({
+    page,
+  }) => {
+    const filterModeSelect = page.getByLabel('Filter mode');
+    await expect(filterModeSelect).toBeVisible();
+    await expect(filterModeSelect).toBeDisabled();
+
+    const abandonedChip = page.getByRole('button', {
+      name: 'Abandoned bicycle',
+      exact: true,
+    });
+    await abandonedChip.click();
+    await expect(filterModeSelect).toBeEnabled();
+
+    // deselecting the only selected chip disables the control again
+    await abandonedChip.click();
+    await expect(filterModeSelect).toBeDisabled();
+  });
+
+  test('chip click toggles its pressed state', async ({page}) => {
+    const abandonedChip = page.getByRole('button', {
+      name: 'Abandoned bicycle',
+      exact: true,
+    });
+
+    await expect(abandonedChip).toHaveAttribute('aria-pressed', 'false');
+    await abandonedChip.click();
+    await expect(abandonedChip).toHaveAttribute('aria-pressed', 'true');
+    await abandonedChip.click();
+    await expect(abandonedChip).toHaveAttribute('aria-pressed', 'false');
+  });
+
+  test('report summary is only marked "(filtered)" once an issue is selected', async ({
+    page,
+  }) => {
+    const summaryLabel = page.getByTestId('report-summary-label');
+    await expect(summaryLabel).not.toContainText('(filtered)');
+
+    await page
+      .getByRole('button', {name: 'Abandoned bicycle', exact: true})
+      .click();
+    await expect(summaryLabel).toContainText('(filtered)');
+
+    await page
+      .getByRole('button', {name: 'Abandoned bicycle', exact: true})
+      .click();
+    await expect(summaryLabel).not.toContainText('(filtered)');
+  });
+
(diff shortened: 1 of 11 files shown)


bikespace_frontend/src/components/dashboard/data-issue-frequency-chart/DataIssueFrequencyChart.tsx
```diff
@@ -22,44 +21,36 @@ function DataIssueFrequencyChart({
...
+        issues.length === 0 || issues.includes(i)
+          ? styles[i]
+          : styles[`${i}_light`],
     }));
 
     setData(inputData);
-  }, [submissions, issue]);
+  }, [submissions, issues]);
 
   useEffect(() => {
```
tallcoleman: I think this effect can be removed entirely; it looks like it was meant to do the analytics tracking for when the chart was used as a filter UI.

bikespace_frontend/src/components/dashboard/filter-issue-type/filter-issue-type.module.scss
```diff
@@ -0,0 +1,75 @@
...
+  display: inline-flex;
+  align-items: center;
+  gap: 6px;
+  height: 1.75rem;
+  padding: 0px 10px;
+  background-color: white;
+  border: 1px solid $color-primary-d50p;
+  border-radius: 6px;
+  font-family: $body-font-stack;
+  font-size: 14px;
```
tallcoleman: Check other font sizes used for consistency

bikespace_frontend/src/components/dashboard/dashboard-page/DashboardPage.tsx
```diff
@@ -87,7 +87,14 @@ export function DashboardPage() {
...
       subs = subs.filter(s => parkingDuration.includes(s.parking_duration));
 
-    if (issue !== null) subs = subs.filter(s => s.issues.includes(issue));
+    if (issues.length !== 0)
+      subs = subs.filter(s => {
+        if (issueFilterMode === 'exclude')
+          return !s.issues.some(i => issues.includes(i));
+        if (issueFilterMode === 'all')
+          return issues.every(i => s.issues.includes(i));
+        return s.issues.some(i => issues.includes(i));
```
tallcoleman, line 98: Please add a comment documenting which user selection corresponds to the default case

bikespace_frontend/src/components/dashboard/dashboard-page/DashboardPage.tsx
```diff
@@ -87,7 +87,14 @@ export function DashboardPage() {
     if (parkingDuration.length !== 0)
       subs = subs.filter(s => parkingDuration.includes(s.parking_duration));
 
-    if (issue !== null) subs = subs.filter(s => s.issues.includes(issue));
+    if (issues.length !== 0)
+      subs = subs.filter(s => {
+        if (issueFilterMode === 'exclude')
```
tallcoleman: Should use the IssueFilterMode Enum here?

bikespace_frontend/src/components/dashboard/filter-issue-type/FilterIssueType.tsx
```diff
@@ -0,0 +1,117 @@
...
+
+  const handleChipClick = useCallback(
+    (value: IssueType) => {
+      const newIssues = issues.includes(value)
+        ? issues.filter(v => v !== value)
+        : [...issues, value];
+
+      setFilters({issues: newIssues});
+
+      trackUmamiEvent('issuefilter', {[value]: value});
```
tallcoleman: Could this track the filter mode as well?

bikespace_frontend/src/components/dashboard/filter-issue-type/FilterIssueType.tsx
```diff
@@ -0,0 +1,117 @@
...
+};
+
+export function FilterIssueType() {
+  const {issues, issueFilterMode, setFilters} = useStore(state => ({
+    issues: state.filters.issues,
+    issueFilterMode: state.filters.issueFilterMode,
+    setFilters: state.setFilters,
+  }));
+
+  const handleChipClick = useCallback(
```
tallcoleman: Is the useCallBack wrapper actually needed here? I didn't notice much of a difference with or without it; though I don't really know much about how to do proper React profiling...

bikespace_frontend/src/components/dashboard/filter-issue-type/FilterIssueType.tsx
```diff
@@ -0,0 +1,117 @@
...
+              <img
+                src={issueIcons[value].src}
+                alt=""
+                className={styles.chipIcon}
+              />
+              {label}
+            </button>
+          ))}
+        </div>
+        {issues.length > 0 && (
```
tallcoleman: Could the filter mode control show even when no chip is selected? That way there would be fewer layout moves and greater discoverability of the control.

bikespace_frontend/src/components/dashboard/filter-issue-type/FilterIssueType.tsx
```diff
@@ -0,0 +1,117 @@
...
+              id="issue-filter-mode"
+              className={styles.filterModeSelect}
+              value={issueFilterMode}
+              onChange={e =>
+                setFilters({
+                  issueFilterMode: e.target.value as IssueFilterMode,
+                })
+              }
+            >
+              <option value={IssueFilterMode.Any}>Match any</option>
```
tallcoleman: Suggestion: add "selected" to each label, e.g. "Match any selected"

bikespace_frontend/src/components/dashboard/dashboard-page/DashboardPage.tsx
tallcoleman, line 1: The green summary box at the top is showing "filtered" even when no filter is applied - line 84 of ReportSummary.tsx needs to be updated to properly check if the issues filter is in the null state.

bikespace_frontend/src/components/dashboard/filter-issue-type/filter-issue-type.module.scss
```diff
@@ -0,0 +1,75 @@
...
+  }
+}
+
+.chipIcon {
+  width: 16px;
+  height: 16px;
+  object-fit: contain;
+}
+
+.chipSelected {
```
tallcoleman, line 41: The design looked like it had:
- unselected: 1px border, white background
- selected: 2px border, $color-primary-10p background

This design is good too, but I'm curious about why you selected it. Better text contrast?

bikespace_frontend/src/components/dashboard/filter-issue-type/filter-issue-type.module.scss
tallcoleman, line 1: For consistency, when we've needed slightly smaller text for UI elements elsewhere, I think we've used `0.9rem` instead of specific pixel values.

bikespace_frontend/src/components/dashboard/filter-issue-type/FilterIssueType.tsx
```diff
@@ -0,0 +1,117 @@
...
+                className={styles.chipIcon}
+              />
+              {label}
+            </button>
+          ))}
+        </div>
+        {issues.length > 0 && (
+          <div className={styles.filterModeRow}>
+            <label htmlFor="issue-filter-mode">Filter mode</label>
+            <select
```
tallcoleman: This could use the SidebarSelect component for consistency

tallcoleman (changes requested):
Thanks @KarishmaChanglani this is really great work! I've left some comments for you to look at; if you want to talk through any of them more, just let me know.

Would you also be willing to write a test that covers the expected calls to state.setFilters? Cases should cover each filter mode, the user selecting one issue type, and the user selecting multiple issue types. `FilterDateRangeCustom.test.tsx` is probably a helpful example to look at for this.

Thanks again - this will be so much better for the users than hidden filtering functionality in the issue chart.

tallcoleman:
@KarishmaChanglani I also made a pending update in #410 to cache playwright browsers which should speed up the end to end testing step and prevent it from timing out. There's still a little bit of flakiness you'll see on the e2e tests from time to time, which may not be fixable until we move the dashboard page over to react-map-gl/maplibre instead of react-leaflet.

bikespace_frontend/src/components/dashboard/data-issue-frequency-chart/DataIssueFrequencyChart.tsx
KarishmaChanglani: Fair point, I was hesitant to just remove it but it makes sense I will clean it up.

bikespace_frontend/src/components/dashboard/filter-issue-type/FilterIssueType.tsx
KarishmaChanglani: I thought it would clutter the view. In hindsight you are right though

bikespace_frontend/src/components/dashboard/filter-issue-type/FilterIssueType.tsx
KarishmaChanglani: I noticed it was way slower without it.

bikespace_frontend/src/components/dashboard/filter-issue-type/filter-issue-type.module.scss
KarishmaChanglani: I'll be honest I didn't notice the difference between the figma and my local until you pointed it out. Will fix.

bikespace_frontend/src/components/dashboard/filter-issue-type/FilterIssueType.tsx
KarishmaChanglani: I am not noticing the slow down on the preview here so I am removing it

bikespace_frontend/src/components/dashboard/filter-issue-type/FilterIssueType.tsx
KarishmaChanglani: The sidebar select will make the drop down show up below the tag. doesn't match the figma perfectly but looks pretty clean to me if you are ok with that

bikespace_frontend/e2etests/filter-issue-type.spec.ts
tallcoleman, line 1: This is a really well written test suite - love the helper functions!

bikespace_frontend/src/components/dashboard/filter-issue-type/FilterIssueType.tsx
```diff
@@ -0,0 +1,120 @@
...
+          ))}
+        </div>
+        <div className={styles.filterModeRow}>
+          <label htmlFor="issue-filter-mode" className={styles.filterModeLabel}>
+            Filter mode
+          </label>
+          <SidebarSelect
+            id="issue-filter-mode"
+            value={issueFilterMode}
+            disabled={issues.length === 0}
```
tallcoleman, line 78: I think it's fine for the select here to be enabled even if the user hasn't selected any chips. Unless it makes something bad happen with the filtering logic?

bikespace_frontend/src/components/dashboard/filter-issue-type/filter-issue-type.module.scss
```diff
@@ -0,0 +1,60 @@
...
```
tallcoleman, line 41: Do you think it looks better to have `gap: 4px;` here (to override the `gap: 6px;` from `.chip`) to offset the layout shift with the border getting thicker?

tallcoleman (approved):
Thanks for going through all the comments and for adding a great test suite! I left a couple optional things you can look at, but up to you whether you think they need tweaking.

PR is approved - please go ahead and merge into main; make sure to squash merge and edit the commit message if needed.

bikespace_frontend/src/components/dashboard/filter-issue-type/filter-issue-type.module.scss
KarishmaChanglani: I tried it and I do think it did, but I also was seeing some weirdness when going from mobile mode to desktop so I didn't add it for now

bikespace_frontend/src/components/dashboard/filter-issue-type/FilterIssueType.tsx
KarishmaChanglani: For me it was showing no results if you pick the exclude option, which... not sure why you would to be fair without picking filters

PR #410, opened 2026-06-04
Improve e2e CI job runtime with Playwright browser caching

tallcoleman:
## Summary

- Cache Playwright browser binaries (keyed on `@playwright/test` version) so the ~300–500 MB download is skipped on repeat runs against the same PR
- OS system libraries (`install-deps`) still run on every job since the runner is ephemeral and they're not cached
- Pin Node.js to `24` (active LTS, supported through April 2028) across all frontend jobs; was `lts/*` which can break silently on LTS transitions
- Pin Postgres service image to `postgres:15` to match production; was untagged `latest`
- Remove `edited` from the workflow trigger types — PR description/title edits don't change code and shouldn't re-run CI
- Add `.github/workflows/test-preview.yml` to the workflow's path filter so changes to the workflow itself trigger a run

## Test plan

- [x] Confirm `test-e2e` passes on first run (cache miss — full browser install)
- [x] Confirm `test-e2e` passes on a second run against this PR (cache hit — browser install step skipped)
- [x] Confirm `test-frontend` and `publish` jobs pass with Node 24
- [x] Confirm playwright-report artifact is still uploaded

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.github/workflows/test-preview.yml
+++ b/.github/workflows/test-preview.yml
@@ -20,7 +20,7 @@ jobs:
       # Label used to access the service container
       postgres:
         # Docker Hub image
-        image: postgres:16
+        image: postgres:15
         # Provide the password for postgres
         env:
           POSTGRES_PASSWORD: postgres
@@ -59,7 +59,7 @@ jobs:
       # Label used to access the service container
       postgres:
         # Docker Hub image
-        image: postgres:16
+        image: postgres:15
         # Provide the password for postgres
         env:
           POSTGRES_PASSWORD: postgres
@@ -112,7 +112,7 @@ jobs:
       # Label used to access the service container
       postgres:
         # Docker Hub image
-        image: postgres:16
+        image: postgres:15
         # Provide the password for postgres
         env:
           POSTGRES_PASSWORD: postgres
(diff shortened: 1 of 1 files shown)


.github/workflows/test-preview.yml
```diff
@@ -19,7 +20,7 @@ jobs:
       # Label used to access the service container
       postgres:
         # Docker Hub image
-        image: postgres
+        image: postgres:15
```
KarishmaChanglani, line 23: I am curious why 15 and not 18? Wouldn't this be us downgrading from 18 if the docker was pulling the latest by default?

tallcoleman: It matches the version currently in production. I should check with Sonal about whether we can upgrade to a newer version though...

KarishmaChanglani: Huh, in my local it was showing as 18 which is a good reason to add this

KarishmaChanglani (approved):

.github/workflows/test-preview.yml
sonalranjit: I think had to choose postgres 15 because that was the supported one on fly.io. But we can look into upgrading it. Even the current one we have on fly.io is considered "legacy postgres"
Screenshot 2026-07-10 at 10 02 02 AM

We could also considered alternative to postgres as well like duckdb, since our db is soo small.

Issue #405, opened 2026-05-18
Enable a login flow via the front-end and and show user account on submissions


tallcoleman:
Undertake after #396 
Enabling step for #382 

## Description

#396 added token authentication to the API, added endpoints to edit or delete submissions, and added an optional user field that is populated if a submission is made while the user is logged in.

To start testing these features on the front-end, the following things need to be added:

- A login form: a simple way to do this might be to enable a login form component on the `/submission` page if the user adds a URL parameter (e.g. `/submission?login=true`). This would allow for testing without displaying a feature that is not generally available to users.
- The login form should use Tanstack Query to make a mutation request at the login endpoint, get the authentication token, and then save it in local storage. The token should be included in auth-required requests using the `Authentication-Token` header. Research best practices for handling and storing tokens in static front-end apps when implementing this.
- If a user is specified in a submission, the `/dashboard` page should show their username on the submission details
- An end to end test should be added for the login flow, if feasible

## Other details to consider

- How can the user log out?
- How does the front-end handle expired tokens?
- How does the front-end handle if the user tries to do an action without authenticating?
- How should anonymous submissions be shown on `/dashboard`? Is it helpful to show something like "anonymous user" or is fine just to omit the user field if the submission was made anonymously?
- Implementing the endpoints to edit or delete submissions is likely not worth undertaking in this task, unless there is an easy way to implement the edit action by calling e.g. `/submission?login=true&submission_id={existing_submission}` and it is not too much extra scope to implement the delete control on `/dashboard` along with a login modal on that page as well.

PR #428, opened 2026-08-13
Enable a login flow via the front-end and and show user account on submissions

tallcoleman:
Implements #405 

**IMPORTANT:** instruct users to update any .local env files they have for the frontend if they override the API value with the root API path change

## Main changes:

- Adds `/profile` to frontend allow the user to login, view their profile info, and logout
- User account is tied to submissions made via `/submission` while logged in. If the user is currently logged in, their username is shown as part of the submission confirmation screen.
- If a user is specified in a submission, the `/dashboard` page now shows their username on the submission details
- `BIKESPACE_API_URL` now specifies the API root URL and excludes the `/api/v2` suffix, since admin routes are not organized under `/api/v2`

## Details:

- New frontend `useAuthStore` hook in `'@/states/store'` to store user auth token
- New API route `/users/me` in `bikespace_api/bikespace_api/admin/admin_routes.py` to show details of the currently logged-in user and confirm token freshness
- New frontend `userQuery` in `bikespace_frontend/src/hooks/use-user-query.ts` to get user details from `/users/me` and show them in the frontend UI
- submission Summary and ProfileDetails components have error handling in case userQuery is in a failed state
- Updated `bikespace_frontend/src/interfaces/Submission.ts` to include `user` field from API
- Added API tests for new `/users/me` route
- Added an end to end test for the flow of logging in, making a submission while logged in, and then logging out
- Added frontend unit tests focused on handling of API error responses

## Incidental Changes:

- `bikespace_api.seed` now exports `admin_user` and `non_admin_user` dicts to replace hardcoded values in tests; corresponding values for end-to-end tests are saved in a new constants.ts
- Added documentation on how to show a loading state until hydration and how to run frontend unit tests on changed files only
- The relative time description in submission items on `/dashboard` now has the first letter capitalized, using a new text utils function `sentenceCase`
- moved AnimatedEllipses component from GeocoderSearch into shared-ui and added aria role and label for this component
- Fixed some Jest asserts that were improperly written and would not catch failure state
- Various styling and code cleanup changes


bikespace_api/bikespace_api/admin/tests/test_admin_routes.py
tallcoleman, line 1: Should add a test to confirm that no details are returned when no user is logged in

bikespace_frontend/src/components/auth/login-form/LoginForm.tsx
tallcoleman, line 1: Add unit tests for the various valid (logged in, logged out) and error states
Actual login POST is handled by end to end tests

bikespace_frontend/src/components/auth/profile-details/ProfileDetails.tsx
tallcoleman, line 1: Add unit tests for the various states displayed

bikespace_frontend/src/components/dashboard/feed-submission-item/FeedSubmissionItem.tsx
tallcoleman, line 1: Add unit tests to cover cases where there is and is not a user included with the submissions data

bikespace_frontend/src/components/dashboard/map-popup/MapPopup.tsx
tallcoleman, line 1: May ignore unit tests on this one - needs to be re-worked in a future PR anyway

bikespace_frontend/src/components/submission/summary/Summary.tsx
tallcoleman, line 1: Update unit test for this component to show user logged in state and userQuery error/loading states

bikespace_frontend/src/utils/textUtils.ts
tallcoleman, line 1: these two functions are great candidates for unit test coverage

tallcoleman (commented):
Note areas where additional unit tests should be added

bikespace_frontend/.env.production
```diff
@@ -1,5 +1,5 @@
 # FOR CONFIG ONLY - DO NOT ADD SECRETS TO THIS FILE
-BIKESPACE_API_URL=https://api-dev.bikespace.ca/api/v2
+BIKESPACE_API_URL=https://bikespace-api-test.tallcolemanprojects.ca/ # TODO delete before merge
```
KarishmaChanglani: did you forget to delete this?

bikespace_frontend/src/components/auth/profile-details/ProfileDetails.tsx
```diff
@@ -0,0 +1,56 @@
...
+            {userQuery.data?.email}{' '}
+            <span title={userQuery.data?.confirmed_at}>
+              ({userQuery.data?.confirmed_at ? 'confirmed' : 'unconfirmed'})
+            </span>
+          </dd>
+        </div>
+      </dl>
+      <LoginForm />
+    </div>
+  ) : userQuery.isError ? (
```
KarishmaChanglani, line 42: I think this will render "the LoginForm tag.  But earlier LoginForm gates on token presence, not validity. Tokens expire after an hour right? but stay in localStorage indefinitely? so a user whose token just expired lands here what do they see? I suppose it's a bit of an edge case but I was curious if they just end up on a dead page with no way to move forward?

bikespace_frontend/src/components/auth/login-form/LoginForm.tsx
```diff
@@ -0,0 +1,138 @@
...
+      );
+      const responseData = await response.json();
+
+      if (response.ok) {
+        // save auth token on successful response
+        setAuthToken(responseData.response.user.authentication_token);
+      } else {
+        // display form error(s) from API validation error response
+        for (const serverError of Object.entries(
+          responseData?.response?.field_errors
```
KarishmaChanglani: if login fails for a reason that isn't a field-validation error like a server error this code might crash instead of showing the real error message. If we want that thats fine but I wanted to point it out.

bikespace_frontend/e2etests/submission.spec.ts
```diff
@@ -94,3 +93,105 @@ test('Submit an issue', async ({page}, testInfo) => {
 
   await expect(page.getByText(testComment).first()).toBeVisible();
 });
+
+test('Log in, submit an issue, log out', async ({page}, testInfo) => {
+  const nonAdminUser = {
+    username: 'nonadminuser',
```
KarishmaChanglani: don't we already have this constant defined earlier. Is this dupe?

KarishmaChanglani (changes requested):

bikespace_frontend/.env.production
tallcoleman: I added it to test the preview with a preview deployment of the API as well. I'll make sure to remove it later!

I was trying to see if there's an easy way to have the PR previews link out to a separate preview API instance instead of the production one. So far I haven't found a super clean way to do it, but I'll keep looking. At the very least, it might be possible to have all the frontend previews connect to a test instance of the API by looking at the `GITHUB_REF_NAME` env var in the github action.

bikespace_frontend/src/components/auth/profile-details/ProfileDetails.tsx
tallcoleman: I think I just ran into this case this morning while showing the work in progress to someone else; looks like it gives a non-informative 401 error and then the user would have to log out and back in again to get a current token. I'll see if I can improve the handling of this case and add a test case for it.

bikespace_frontend/src/components/auth/login-form/LoginForm.tsx
tallcoleman: In this case I'm assuming that if the response is in the 200 range, the API will have returned a properly formed response that can be parsed. If there's a server error, I think it would land in the catch block on line 64 instead?

bikespace_frontend/e2etests/submission.spec.ts
tallcoleman: It is a duplication, but it's because the other constant is defined in the API codebase, which the e2e code can't import and read. I think in the future if we switched the playwright tests to use the python driver instead of the JS one, we would be able to just define it in one place. Maybe for now I should add a comment that this needs to align with the values set in `bikespace_api/bikespace_api/seed.py`?

bikespace_frontend/src/components/auth/profile-details/ProfileDetails.tsx
KarishmaChanglani: Sounds good!

bikespace_frontend/src/components/auth/login-form/LoginForm.tsx
```diff
@@ -0,0 +1,138 @@
...
+    errors.email ? {key: 'email', message: errors.email?.message} : null,
+    errors.password
+      ? {key: 'password', message: errors.password?.message}
+      : null,
+    errors.root?.serverError
+      ? {key: 'root.serverError', message: errors.root?.serverError?.message}
+      : null,
+  ];
+
+  return authToken ? (
```
tallcoleman: Think about whether you want to use useUserQuery here since it would test the token validity before showing the user whether they need to log in again or not. Currently this happens because ProfileDetails depends on useUserQuery and wraps the login form, but since LoginForm is a standalone component it should also handle a revoked or expired token more smoothly itself.

A more elegant solution may be to define a getter function for when the token is retrieved that checks its freshness (see: https://flask-security-too.readthedocs.io/en/stable/features.html#token-authentication ) and returns null if the token is expired. This would still leave the revoked token case unhandled, however.

KarishmaChanglani (approved):

bikespace_frontend/src/components/auth/profile-details/ProfileDetails.tsx
tallcoleman: Should be handled now in https://github.com/bikespace/bikespace/pull/428/commits/4390483d60608334ca5f95f0f850ea4bb23457c4 - if the user query receives a 401 it removes the auth token so that the user is prompted to log in again

bikespace_frontend/.env.development
tallcoleman, line 1: Note for self - should just remove the separate dashboard API URL since there is now behaviour that requires the submission form and dashboard display to pull from the same API instance

Issue #293, opened 2025-10-25
Replace deprecated @import in scss with @use


tallcoleman:
Issue, per the [scss documentation](https://sass-lang.com/documentation/at-rules/import/):

"As of Dart Sass 1.80.0, the `@import` rule is [deprecated](https://sass-lang.com/documentation/breaking-changes/import) and will be removed from the language in Dart Sass 3.0.0. Prefer the [@use rule](https://sass-lang.com/documentation/at-rules/use) instead."

Quick fix in some cases would be to e.g. replace:

```scss
@import '@/styles/variables.module.scss';
```

with:

```scss
@use './styles/variables.module.scss' as *;
```

but it may be worthwhile to embrace the namespacing of imported variables that you get with the recommended usage of `@use`. In which case, we should think about whether there should be a different name for the `variables.module.scss` file so that its namespace is more descriptive.

There is also a [migration CLI tool](https://sass-lang.com/documentation/cli/migrator/) that may make the update easier.

SuganyaMaheswaran:
Hi! I'd love to pick this up and work on replacing the deprecated @import rules with @use. I'll take a look at using the migration CLI tool or updating them manually where appropriate. Please assign this to me if this still available.

PR #432, opened 2026-08-30
Frontend/biketheft-feature

thesun7385:
Added Bike thefts history feature


bikespace_frontend/src/components/content-pages/content-blocks/content-blocks.tsx
```diff
@@ -46,6 +46,13 @@ function HeroBlock({
           >
             Level of Traffic Stress map
           </a>
+          <a
+            href="/stolen-history-map"
+            className={styles.buttonLink}
+            data-umami-event="bike-stolen-map-from-frontpage"
+          >
+            Bike stolen history
```
tallcoleman, line 54: Suggest "Reported bike theft map" or "Bike theft map". And maybe `/bike-theft-map` for the url?

bikespace_frontend/src/components/stolen-history/stolen-map-layout/StolenHistoryMapLayout.tsx
```diff
@@ -0,0 +1,25 @@
+'use client';
+
+import React from 'react';
+import {QueryClientProvider} from '@tanstack/react-query';
+
+import {queryClient} from '@/config/query-client';
+
+import {DashboardHeader} from '@/components/dashboard/dashboard-header';
```
tallcoleman, line 8: (Optional) We need to move this component into a shared UI folder since it's now used in all of the map pages. Ideally, the email feedback mailto link would be configurable so that the template message reflects the specific page that feedback is given from.

bikespace_frontend/src/components/stolen-history/stolen-map-layout/StolenHistoryMapLayout.tsx
```diff
@@ -0,0 +1,25 @@
...
+import styles from './stolen-map-layout.module.scss';
+
+export function StolenHistoryMapLayout({
+  children,
+}: {
+  children: React.ReactNode;
+}) {
+  return (
+    <QueryClientProvider client={queryClient}>
+      <div className={styles.ltsMapLayout}>
```
tallcoleman, line 19: the names for style classes copied over from the LTS map should be updated to match the name of this page

bikespace_frontend/src/components/stolen-history/stolen-map-page/stolen-map-page.module.scss
```diff
@@ -0,0 +1,186 @@
+@use '@/styles/variables.module.scss' as variables;
```
tallcoleman, line 1: Good catch to do this the up-to-date way; gives us less to do for #293 !

bikespace_frontend/src/components/stolen-history/stolen-map-page/stolen-map-page.module.scss
```diff
@@ -0,0 +1,186 @@
...
+  margin: 4px;
+  display: flex;
+  flex-direction: column;
+  gap: 4px;
+  max-height: 220px;
+  overflow-y: auto;
+  word-break: break-word;
+}
+
+.ltsDescriptionList {
```
tallcoleman, line 66: Classes copied over from LTS stylesheet should have the names updated to match this page

bikespace_frontend/src/components/stolen-history/stolen-map-page/stolen-map-page.module.scss
tallcoleman, line 1: As you package out some of the map page code into separate components, make sure to package out the corresponding styles as well.

In addition, some of the filter controls are missing styling (example from Safari on mac):

Image

And sometimes the info cards are rendering with an internal scroll bar:

Image

bikespace_frontend/src/components/stolen-history/stolen-map-page/StolenHistoryMapPage.tsx
```diff
@@ -0,0 +1,606 @@
...
+  MapRef,
+  MapStyle,
+} from 'react-map-gl/maplibre';
+
+// Show the zoom in-out/current locaiton
+import 'maplibre-gl/dist/maplibre-gl.css';
+import styles from './stolen-map-page.module.scss';
+import parkingStyles from '@/components/parking-map/parking-map-page/parking-map-page.module.scss';
+
+// --- PMTiles setup ---
```
tallcoleman, line 31: This chunk can be removed since the page is not using a pmtiles source

bikespace_frontend/src/components/stolen-history/stolen-map-page/StolenHistoryMapPage.tsx
```diff
@@ -0,0 +1,606 @@
...
+  CircleLayer,
+  MapLayerMouseEvent,
+  MapRef,
+  MapStyle,
+} from 'react-map-gl/maplibre';
+
+// Show the zoom in-out/current locaiton
+import 'maplibre-gl/dist/maplibre-gl.css';
+import styles from './stolen-map-page.module.scss';
+import parkingStyles from '@/components/parking-map/parking-map-page/parking-map-page.module.scss';
```
tallcoleman, line 29: (Optional) The styles here that are being used by this map and the LTS map should really be moved into a shared folder so that it's clear that multiple pages depend on it.

bikespace_frontend/src/components/stolen-history/stolen-map-page/StolenHistoryMapPage.tsx
```diff
@@ -0,0 +1,606 @@
...
+      url: 'pmtiles://backup_map/toronto.pmtiles',
+      attribution:
+        '<a href="https://protomaps.com">Protomaps</a> © <a href="https://openstreetmap.org">OpenStreetMap</a>',
+    },
+  },
+  layers: layers('protomaps', namedFlavor('light'), {lang: 'en'}),
+};
+
+// --- Stolen bike report type ---
+export interface StolenBikeReport {
```
tallcoleman, line 57: Should this be placed in `bikespace_frontend/src/interfaces/BikeTheftProperties.tsx` and imported from there?

bikespace_frontend/src/components/stolen-history/stolen-map-page/StolenHistoryMapPage.tsx
```diff
@@ -0,0 +1,606 @@
...
+  id: string;
+  date: string; // ISO date string e.g. "2024-03-15"
```
tallcoleman, line 59: The comments here describing what the strings usually look like/include are very helpful!

bikespace_frontend/src/components/stolen-history/stolen-map-page/StolenHistoryMapPage.tsx
```diff
@@ -0,0 +1,606 @@
...
+  location: string; // Human-readable address or area
+  bikeType: string; // e.g. "Road bike", "Mountain bike"
+  color: string;
+  description: string;
+  status: 'stolen' | 'recovered';
+  latitude: number;
+  longitude: number;
+}
+
+// Import stolen bike reports from the repo dataset.
```
tallcoleman, line 69: For production, the data flow should look like:

- in the `bikespace_frontend/.env.*` files, add `DATA_BICYCLE_THEFT` with a value like `https://raw.githubusercontent.com/bikespace/parking-map-data/refs/heads/data/bike_theft/<path_to_file>.geojson`
- add the `DATA_BICYCLE_THEFT` to the list in `next.config.mjs`
- In the component for the bike theft layer, add a react-map-gl/maplibre `<Source>` component with the `data` url being derived from `process.env.DATA_BICYCLE_THEFT`. You can keep the rest of the logic pretty similar to what you have now, but instead of the GeoJSON helper functions, you should just be able to make a copy of the GeoJSON with the `features` list replaced with the output from the filtering.

bikespace_frontend/src/components/stolen-history/stolen-map-page/StolenHistoryMapPage.tsx
```diff
@@ -0,0 +1,606 @@
...
+import stolenBikeReportsData from '../../../../../datasets/stolen-report/stolen_bike_reports.json';
+
+const stolenBikeReports: StolenBikeReport[] =
+  stolenBikeReportsData as StolenBikeReport[];
+
+// --- Map layer for stolen bike pins ---
+const stolenPinsLayer: CircleLayer = {
```
tallcoleman, line 76: Ideally you want to make the map layer an importable component to separate the map layer rendering logic from the page rendering logic. It also opens up the option to easily show the bike theft pins as a context layer on another page (e.g. parking map) by just putting the component into the map layers list.

bikespace_frontend/src/components/stolen-history/stolen-map-page/StolenHistoryMapPage.tsx
```diff
@@ -0,0 +1,606 @@
...
+      stops: [
+        [10, 5],
+        [16, 12],
+      ],
+    },
+    'circle-color': [
+      'match',
+      ['get', 'status'],
+      'recovered',
+      '#2e7d32',
```
tallcoleman, line 92: I would investigate if there is another colour we can use for recovered thefts, since it's hard to distinguish between red and green for users with protanopia (there's one green bubble in the screenshot below, near the middle, a bit to the right):

Image

The contrast for the selected state is better and aided by the growth in size, so the colour used there might be fine, though a "center on screen" button for the selected pin would help if the user is having difficulty locating the selected pin on the map.

Image

bikespace_frontend/src/components/stolen-history/stolen-map-page/StolenHistoryMapPage.tsx
```diff
@@ -0,0 +1,606 @@
...
+    },
+    'circle-color': '#ff6f00',
+    'circle-stroke-width': 3,
+    'circle-stroke-color': '#ffffff',
+    'circle-opacity': 0.9,
+  },
+};
+
+// --- GeoJSON helpers ---
+function reportsToGeoJSON(reports: StolenBikeReport[]) {
```
tallcoleman, line 132: These helpers can probably be eliminated if the source data is already .geojson

bikespace_frontend/src/components/stolen-history/stolen-map-page/StolenHistoryMapPage.tsx
```diff
@@ -0,0 +1,606 @@
...
+        padding: '2px 10px',
+        borderRadius: 12,
+        fontSize: '0.8rem',
+        fontWeight: 600,
+        backgroundColor: isRecovered ? '#e8f5e9' : '#ffebee',
+        color: isRecovered ? '#2e7d32' : '#c62828',
+        border: `1px solid ${isRecovered ? '#a5d6a7' : '#ef9a9a'}`,
+      }}
+    >
+      {isRecovered ? '✓ Recovered' : '✗ Stolen'}
```
tallcoleman, line 186: The checkmark and cross mark should be wrapped in something like `<span aria-hidden="true">` so that they are not read by screen readers. Especially the cross mark gets read as "ballot x" which is potentially confusing. When I tested on VoiceOver, it skipped them by default unless reading character by character, but other screen readers may read these characters out loud.

bikespace_frontend/src/components/stolen-history/stolen-map-page/StolenHistoryMapPage.tsx
```diff
@@ -0,0 +1,606 @@
...
+
+// --- Main component ---
+export function StolenHistoryMapPage() {
+  const [defaultLocation, setDefaultLocation] = useState(defaultMapCenter);
+  const [selectedReport, setSelectedReport] = useState<StolenBikeReport | null>(
+    null
+  );
+  const [sidebarIsOpen, setSidebarIsOpen] = useState(true);
+  const [geoSearchIsMinimized, setGeoSearchIsMinimized] = useState(false);
+  const [statusFilter, setStatusFilter] = useState<
```
tallcoleman, line 199: Should these three options be handled as an enum?

bikespace_frontend/src/components/stolen-history/stolen-map-page/StolenHistoryMapPage.tsx
```diff
@@ -0,0 +1,606 @@
...
+    'all' | 'stolen' | 'recovered'
+  >('all');
+
+  // Year-month-date filters and set current year and month as default
+  const [yearFilter, setYearFilter] = useState<string>(
```
tallcoleman, line 204: You can leave out of scope for this PR if you want, but I think the ideal filtering would be to give the user a date range option, similar to what we have in the submission dashboard page.

If you want to do a range option but avoid the scope of all the detailed date handling in this PR it could just be a year range with a from year and a to year in case the user wants to look at multiple years of data (e.g. the last 3 years) at once.

bikespace_frontend/src/components/stolen-history/stolen-map-page/StolenHistoryMapPage.tsx
```diff
@@ -0,0 +1,606 @@
...
+    String(new Date().getFullYear())
```
tallcoleman, line 205: If the user visits this page in January 2027 and the data was last updated in June 2026, what happens here?

bikespace_frontend/src/components/stolen-history/stolen-map-page/StolenHistoryMapPage.tsx
```diff
@@ -0,0 +1,606 @@
...
+  );
+  const [monthFilter, setMonthFilter] = useState<string>('all');
+  const [dayFilter, setDayFilter] = useState<string>('all');
+
+  // Location filter
+  const [locationFilter, setLocationFilter] = useState('all'); // Default to "Outdoor" as per the dataset
+
+  // --- Derive available filter options from data ---
```
tallcoleman, line 213: Your comments here are very helpful in following the logic - thank you!

bikespace_frontend/src/components/stolen-history/stolen-map-page/StolenHistoryMapPage.tsx
```diff
@@ -0,0 +1,606 @@
...
+  const availableYears = React.useMemo(() => {
```
tallcoleman: Remind me to get your thoughts on pre-computing filter values on the data pipeline side. I was thinking it might help make some of the UI rendering for filters on the front-end side easier, especially in cases where you want to rely on pulling the data from the map using e.g. `.queryRenderedFeatures`.

bikespace_frontend/src/components/stolen-history/stolen-map-page/StolenHistoryMapPage.tsx
```diff
@@ -0,0 +1,606 @@
...
+    'June',
+    'July',
+    'August',
+    'September',
+    'October',
+    'November',
+    'December',
+  ];
+
+  const filteredReports = React.useMemo(() => {
```
tallcoleman, line 262: I like your approach here where the filtering is done in JS but then the points are all rendered as a GeoJSON source. Gives you the performance benefit of having MapLibre render GeoJSON natively without the complexity of dealing with MapLibre filter expressions. I'll probably want to switch the parking map over to working this way!

bikespace_frontend/src/components/stolen-history/stolen-map-page/StolenHistoryMapPage.tsx
```diff
@@ -0,0 +1,606 @@
...
+  const mapStyleRoadLabelsLayer = process.env.MAPTILER_API_KEY
+    ? 'Road labels'
+    : 'roads_labels_major';
+
+  const allPinsGeoJSON = reportsToGeoJSON(filteredReports);
+  const selectedPinGeoJSON = selectedReport
+    ? reportToSelectedGeoJSON(selectedReport)
+    : null;
+
+  useEffect(() => {
```
tallcoleman, line 308: This can be removed since the page is not using a pmtiles source

bikespace_frontend/src/interfaces/BikeTheftProperties.tsx
```diff
@@ -0,0 +1,16 @@
+export interface BikeTheftProperties {
```
tallcoleman, line 1: It looks like this interface is not being used anywhere?

datasets/stolen-report/bicycle-thefts-etl.py
tallcoleman, line 1: This should be removed from this PR since it is going into the other repo

datasets/stolen-report/bicycle-thefts.geojson
tallcoleman, line 1: This should be removed from this PR since it is going into the other repo

datasets/stolen-report/stolen_bike_reports.json
tallcoleman, line 1: This should be removed from this PR since it is going into the other repo. If you want to pull in a static file to render for dev purposes, I would suggest moving this to a new folder in `bikespace_frontend/public` and importing from there?

bikespace_frontend/src/components/stolen-history/stolen-map-page/StolenHistoryMapPage.tsx
```diff
@@ -0,0 +1,606 @@
+'use client';
+import React, {useEffect, useRef, useState} from 'react';
+import Map, {
+  GeolocateControl,
+  Layer,
+  NavigationControl,
+  Source,
+} from 'react-map-gl/maplibre';
+import maplibregl from 'maplibre-gl';
+import {Protocol} from 'pmtiles';
```
tallcoleman, line 10: This Protocol import can be removed since this page is not loading a pmtiles source

bikespace_frontend/src/components/stolen-history/stolen-map-page/StolenHistoryMapPage.tsx
```diff
@@ -0,0 +1,606 @@
...
+      error => {
+        if (process.env.NODE_ENV !== 'production') {
+          console.warn('Geolocation failed:', error.message);
+        }
+      },
+      {enableHighAccuracy: true, timeout: 10000}
+    );
+  }, []);
+
+  function handleMapClick(event: MapLayerMouseEvent) {
```
tallcoleman, line 336: There are actually more than one theft points for some intersections, so the handling here should be more like the bike parking map, where it shows a list of all the clicked features. An example would be Kipling & Finch, which has 3 reports in 2025:

Image

Unlike the bike parking map, you may not need to make a distinction between the list of reports in the initial click area and the specific report the user wants to select since the most likely case is that the click returns several points with the exact same geometry. Once you've implemented all the click results, try clicking on a point when zoomed out a lot to see whether additional handling is needed to help the user cross-reference a listed report and the points on the map.

bikespace_frontend/src/components/stolen-history/stolen-map-page/StolenHistoryMapPage.tsx
```diff
@@ -0,0 +1,606 @@
...
+  function handleClearSelection() {
+    setSelectedReport(null);
+    setGeoSearchIsMinimized(false);
+  }
+
+  return (
+    <main className={styles.stolenHistoryMapPage}>
+      <Sidebar isOpen={sidebarIsOpen} setIsOpen={setSidebarIsOpen}>
+        <div className={parkingStyles.sideBarContainer}>
+          {/* --- Report detail card --- */}
```
tallcoleman, line 364: The report detail card should be a separate component

bikespace_frontend/src/components/stolen-history/stolen-map-page/StolenHistoryMapPage.tsx
```diff
@@ -0,0 +1,606 @@
...
+          <GeocoderSearch
+            mapRef={mapRef}
+            isMinimized={geoSearchIsMinimized}
+            setIsMinimized={setGeoSearchIsMinimized}
+            selectResultEvent="stolen-map-select-geosearch-result"
+            clearSearchEvent="stolen-map-clear-geosearch"
+            geosearchErrorEvent="stolen-map-geosearch-error"
+          />
+
+          {/* --- Filters --- */}
```
tallcoleman, line 414: The filter box should probably be a separate component and you may even want to break some of the specific filter controls into components of their own

bikespace_frontend/src/components/stolen-history/stolen-map-page/StolenHistoryMapPage.tsx
```diff
@@ -0,0 +1,606 @@
...
+                  Showing <strong>{filteredReports.length}</strong> report
+                  {filteredReports.length !== 1 ? 's' : ''}
+                </p>
+
+                <p style={{marginTop: 10}}></p>
+              </div>
+            </SidebarDetailsContent>
+          </SidebarDetailsDisclosure>
+
+          {/* --- Legend --- */}
```
tallcoleman, line 530: Legend should be a separate component

bikespace_frontend/src/components/stolen-history/stolen-map-page/StolenHistoryMapPage.tsx
tallcoleman, line 1: This may be more of a data pipeline problem, but the points with no lon/lat are rendering on [null island](https://en.wikipedia.org/wiki/Null_Island):

Image

tallcoleman (changes requested):
I'm really excited for when we get to put this on the site. Even though I've looked at this bike theft data before, I had some new observations from seeing it on an interactive map like this. It'll be so interesting to put it in front of users and see how they interpret and interact with it.

Here are the main things I would suggest you think about to get this ready for production:

## Naming

It might be helpful to run a quick poll in the channel to workshop the page name and url, e.g. “Reported bike theft map”, “Bike theft map”, “Stolen bike map”, etc.

## Structure

I found the code very easy to follow and your structure and commenting is already very good. That being said, there’s a lot of display code and logic in StolenHistoryMapPage.tsx that should be broken up into separate files and components. I would suggest something like:

- Create `@/components/map-layers/bike-theft/` to hold:
    - Bike theft map layer component
    - Bike theft legend component
    - Component for showing the details of an individual theft report
    - Constants for working with the layer (e.g. names of interactive or hit layers)
- In src/components/stolen-history/stolen-map-page/ add additional files for:
    - Filter UI component(s)

The goal would be that you could e.g. import the map layer component into the bike parking map, drop it in, and it would just render correctly.

We’ve also ended up reusing a lot of code from other pages in this implementation as well as in the LTS map. Ideally if a component or set of styles is being used by multiple pages it should be in a shared folder instead of page-specific so that it’s clear when updating it that any changes need to be tested or checked on multiple pages. Some of these re-uses were already in other parts of the code before your PR, so you can decide how much time you want to spend addressing this, and anything left after this PR can be worked on as a separate issue.

## Point Click Handling

In many cases, there will be more than one reported bike theft per intersection, so the click handling should return a list of theft reports and the sidebar should render them all.

## Data Loading

Once the data pipeline is set up, we will want to load the data from a github raw content link passed via environment variable like we do with the parking map. I’ve left a comment with the key steps involved.

## Future Work / Other Ideas

- It would be so cool to be able to pick different variables for the point rendering (e.g. color code by location type, etc.)
- We should see if we can get someone to write a short info blurb (like we have on the LTS map) to cover some key points such as the fact that the data is only updated annually, the geometry being to the nearest intersection, and the research on the percentage of bike thefts that go unreported
- If we have time, it would be interesting to try a more detailed basemap and/or adding the bike lanes context layer to see if that helps users orient themselves to the location of the thefts happening

bikespace_frontend/src/components/stolen-history/stolen-map-page/StolenHistoryMapPage.tsx
thesun7385: Um, this is interesting. I forgot to think that even though the report_ids are unique, they can still point to the same location. This issue needs to be addressed. Great catch! The best way to handle is to show all report_ids together if they have the same stolen location.

tallcoleman:
@thesun7385 the data pipeline run was successful - for reference, the data URL is [`https://raw.githubusercontent.com/bikespace/parking-map-data/refs/heads/data/bike_thefts/stolen_bike_reports.geojson`](https://raw.githubusercontent.com/bikespace/parking-map-data/refs/heads/data/bike_thefts/stolen_bike_reports.geojson)

bikespace_frontend/src/components/stolen-history/stolen-map-page/stolen-map-page.module.scss
thesun7385: TY, I forgot about that. Let me use the same style as the bike parking map for consistency.

bikespace_frontend/src/components/stolen-history/stolen-map-page/StolenHistoryMapPage.tsx
thesun7385: This is a really good idea. I'll include that in the filter.

bikespace_frontend/src/components/stolen-history/stolen-map-page/StolenHistoryMapPage.tsx
thesun7385: Thanks for spotting this. It’s clearly an accessibility issue. Let me figure out how to handle it.

datasets/stolen-report/stolen_bike_reports.json
thesun7385: Oh I got it! I was previously little bit confused as I thought you guys had the datasets dir for reference.

bikespace_frontend/src/components/stolen-history/stolen-map-page/StolenHistoryMapPage.tsx
thesun7385: Totally agree in order to improve readability and consistency.

bikespace_frontend/src/components/stolen-history/stolen-map-page/StolenHistoryMapPage.tsx
thesun7385: This is clearly a pipeline issue. I will investigate this in the pipeline later as it requires some work to check the data consistency.

thesun7385:
### Updated Bike Theft Feature (19/09/2026)

Thank you for the helpful feedback @tallcoleman. It helped me a lot for this commit.

[View the lastest commit](https://github.com/bikespace/bikespace/pull/432/commits/661c2ae14befe2ad09ba2c21bca68150ae1e4a9a)

Here what I have updated:

1. Renamed the bike theft directories and components for consistency.
2. Refactored and updated side-bar components to improve readability.
3. Created 4 separate components for reusability:
- `BikeTheftReportCard`
- `BikeTheftGeocoderSearch`
- `BikeTheftFiltersPanel`
- `BikeTheftLegendPanel`
4. Replaced the local JSON import with fetching from the hosted GeoJSON dataset.
5. Added map clustering with colours (stolen, recover, mix-status reports)
6. Added report counts, date-range filters with presets buttons and support for viewing multiple reports at the same locaions.
Screenshot 2026-09-19 at 3 25 47 PM
Screenshot 2026-09-19 at 3 44 25 PM

### Proposed follow-up work
I would like to address the following in a separate PR as I need more input from your guys. That way, I can fix these items along with in the Future Work / Other Ideas section.
- Color accessibility: Improve the distinction between stolen and recovered for user with red-green colour bliness.(https://github.com/bikespace/bikespace/pull/432#discussion_r3940440801)
- Missing or invalid coordinates: Verify data handling and prevent reports without locations (https://github.com/bikespace/bikespace/pull/432#discussion_r3940773457)

PR #433, opened 2026-08-30
Fix/parking time tz aware

### LINKED ISSUE #391: Make all datetimes used by the API and frontend timezone-aware
Currently, the following date-times in the API are not timezone-aware:
- `User.confirmed_at` (based on `flask_security.core.UserMixin`)
- `Submission.parking_time`

One date-time _is_ timezone-aware: `Submission.submitted_datetime`.

The `Submission.parking_time` in practice is saved in UTC, but its return value being timezone-naive means that additional work has to be done on the front-end to render it correctly. There are currently 11 instances of the string `'+00:00'` being appended to `Submission.parking_time` in order to account for this.

Suggested approach:

- Change `Submission.parking_time` to be timezone-aware (update API and frontend code, generate migration)
- Investigate whether it makes sense to make `User.confirmed_at` timezone-aware or not (since it comes from flask_security).
- Update [documentation for dashboard page](https://docs.google.com/document/d/1zytoCy9mCvxrX4fz4Y1D9DuTs5zaua7C5XL1JqtVMD0/edit?usp=sharing) and save new PDF of updated doc to location in frontend public assets.

neal-bpm:
## Summary

  Makes `parking_time` timezone-aware across the stack (database → API →frontend). It was previously stored and    serialized without an offset, so the frontend had to assume UTC by appending `'+00:00'`. This PR carries the
timezone through the API contract and drops that workaround.

  ## Changes

  **API**
  - `parking_time` column: `DateTime` → `DateTime(timezone=True)`.
  - Marshmallow field: `DateTime` → `AwareDateTime` (ISO format), so responses carry a UTC offset and input           requires one.
  - Alembic migration `dea75e305cf6` alters the type on `bikeparking_submissions` and `bikeparking_submissions_version` (with downgrade).
  - Route tests send ISO 8601 strings with offset and assert via `datetime.fromisoformat`.

  **Frontend**
  - Removed the redundant `+ '+00:00'` when parsing `parking_time` across the dashboard components and the `useAllSubmissionsDateRange` hook.
  - `FilterDateRangeCustom`: derive "today" from `DateTime.local().toISODate()` instead of UTC `toISOString()`, and parse picker values in the local zone, so ranges land on the user's local day (fixes off-by-one near midnight ET).
  - Test mocks updated to offset-bearing timestamps.

  ## Testing
  - `make test-api`
  - `TZ='America/Toronto' make test-frontend`

  ## Migration
  Run `make upgrade-db` (revision `dea75e305cf6`). Existing timestamps are read as UTC after the change.

  ## Issue
https://github.com/bikespace/bikespace/issues/391


bikespace_api/migrations/versions/dea75e305cf6_.py
```diff
@@ -0,0 +1,52 @@
+"""empty message
```
tallcoleman: Please replace "empty message" with a short description of what the migration changes

bikespace_frontend/src/components/dashboard/filter-date-range-custom/FilterDateRangeCustom.test.tsx
```diff
@@ -7,7 +7,7 @@ import {DateRangeInterval} from '@/interfaces/Submission';
 import {FilterDateRangeCustom} from './FilterDateRangeCustom';
 
 // Mock timing to set failure state where the date corresponding to the current time differs between UTC and ET
-const todayDate = new Date('2024-12-31T22:00:00');
```
tallcoleman: Need to understand the rationale for this change

bikespace_frontend/src/components/dashboard/filter-date-range-custom/FilterDateRangeCustom.tsx
```diff
@@ -32,12 +31,14 @@ export function FilterDateRangeCustom() {
   const onSubmit = (data: CustomDateRangeSchema) => {
     setFilters({
       dateRange: {
```
tallcoleman, line 30: My understanding of what's happening here:

- the date HTML input stores its value as a 'YYYY-MM-DD' string
- react-hook-form is set with `{valueAsDate: true}`, which means it calls the JS `Date` constructor on the 'YYYY-MM-DD' string value.
- The JS `Date` constructor assumes the timezone for 'YYYY-MM-DD' strings is UTC
- The `data.from` value is then the YYYY-MM-DD date in UTC, NOT the YYYY-MM-DD date in the user's timezone as intended
- Therefore, you have to get the parsed date's 'YYYY-MM-DD' string portion and re-interpret it as being in the user's timezone

If that matches with your understanding, what happens if we remove `{valueAsDate: true}` from the react-hook-form register function? Can the date parsing here be simplified then to `DateTime.fromISO(data.from, {zone: 'local',})`?

bikespace_api/migrations/versions/dea75e305cf6_.py
tallcoleman, line 1: When I ran `make dev-api` it included the following error as part of the migration run at startup. Do you also get this error? It seems strange to me since we have other migration steps that specify autoincrement, and they don't throw this error.

```log
bikespace_api-1  | /usr/local/lib/python3.12/site-packages/alembic/ddl/postgresql.py:177: UserWarning: autoincrement and existing_autoincrement only make sense for MySQL
bikespace_api-1  |   super().alter_column(
```

bikespace_api/bikespace_api/submissions/submissions_models.py
tallcoleman: General comment - move elsewhere

Testing with this branch - parking time submits as:
`2026-09-01T22:17:29.810Z`

Main branch does the exact same.

Is the submission form updated also? Does the API handle the 'Z' suffix OK? Should it be tested to confirm it handles `+00:00` or another tz suffix? Should line 47 on Submission.tsx be made more specific? Also look at onSubmit in SubmissionForm.tsx and trace back from there to confirm that the timezone handling on the frontend side is still being handled correctly.

bikespace_frontend/src/components/dashboard/filter-date-range-custom/FilterDateRangeCustom.tsx
```diff
@@ -18,8 +18,7 @@ export function FilterDateRangeCustom() {
     setFilters: state.setFilters,
   }));
 
-  const today = new Date();
-  const todayDate = today.toISOString().substring(0, 10);
+  const todayDate = DateTime.local().toISODate();
```
tallcoleman, line 21: You actually caught a bug here, thank you!

`.toISOString()` outputs a date in the UTC time zone, whereas the intended behaviour is for the set date to be in the user's time zone. Your use of `DateTime.local()` fixes it.

bikespace_frontend/src/components/dashboard/filter-date-range-custom/FilterDateRangeCustom.test.tsx
```diff
@@ -7,7 +7,7 @@ import {DateRangeInterval} from '@/interfaces/Submission';
...
+const todayDate = new Date('2025-01-01T00:00:00');
```
tallcoleman: The issue with this test is not the mock date, the issue is that it uses the same faulty logic to test the correct condition on line 52 as the production code. Please revert the change on this line and see the comment on line 52.

It maybe helpful to note in a comment if the existing comment is not clear enough that this mock date is intended to simulate the user using the app (in conjunction with the `TZ='America/Toronto'` jest config) at 22:00 in Toronto time, which would be one date in EST and another date in UTC.

bikespace_frontend/src/components/dashboard/filter-date-range-custom/FilterDateRangeCustom.test.tsx
tallcoleman: This line is the error in the test. It should be:
```ts
const todayDate = '2024-12-31';
```
instead to explicitly test the assumption that if it is Dec 31 in the user's timezone, the UI also shows Dec 31.

tallcoleman (changes requested):
Thank you for working on this! I've left a couple comments for you to look at with some requested changes.

You actually found a bug in the existing code for the custom date filter, and you've fixed the root cause, but the test and form parsing need to be updated in a different way to complete the fix.

I also confirmed that the `/submission` form on the frontend is submitting the parking times e.g. as `2026-09-01T22:17:29.810Z` (for a test submission made around 6pm Toronto time), so the API should be receiving the right input already (i.e. a tz-aware datetime) to align with these changes. 

I'm not sure how good the test coverage is for the frontend submission form part of the codebase since it's from the earlier days of the project, but we might be able to add coverage for it in `bikespace_frontend/e2etests/submission.spec.ts`. If you're not interested in learning how the end to end testing suite works just yet, I can take a look and suggest the changes needed - just let me know.

bikespace_api/migrations/versions/dea75e305cf6_.py
neal-bpm: Added the change comments

bikespace_api/migrations/versions/dea75e305cf6_.py
neal-bpm: I am also seeing the same error in my local environment. Will investigate it.

neal-bpm: autoincrement=False in batch_op.alter_column() triggers a UserWarning from Alembic's PostgreSQL DDL layer ("autoincrement only makes sense for MySQL") because ALTER COLUMN in PostgreSQL has no concept of autoincrement — it's a MySQL-specific DDL feature.

tallcoleman (approved):
Excellent work, thanks again!

I've approved this PR, so feel to squash merge to push it to production. Please make sure to add a descriptive commit message; the summary and changes from your PR description would be perfect if you can also add in a "fixes #391" somewhere in there as well.

neal-bpm:
fix(parking_time): make parking_time timezone-aware across the stack

  Makes parking_time timezone-aware from database through API to frontend.
  It was previously stored and serialised without an offset, forcing the
  frontend to assume UTC by appending '+00:00'.

API:
  - parking_time column: DateTime → DateTime(timezone=True)
  - Marshmallow field: DateTime → AwareDateTime (ISO format), so responses
    carry a UTC offset and input requires one
  - Alembic migration dea75e305cf6 alters the type on
    bikeparking_submissions and bikeparking_submissions_version
  - Route tests send ISO 8601 strings with offset and assert via
    datetime.fromisoformat

  Frontend:
  - Remove redundant '+ +00:00' workaround when parsing parking_time
    across dashboard components and useAllSubmissionsDateRange hook
  - FilterDateRangeCustom: derive "today" from DateTime.local().toISODate()
    instead of UTC toISOString(); parse picker values using setValueAs with
    a local-time suffix so date ranges land on the user's local day, fixing
    off-by-one errors near midnight ET
  - Test mock time updated to trigger the UTC/local date boundary case

  partially fixes #391

PR #435, opened 2026-09-07
Standardize Leaflet map tile providers

### LINKED ISSUE #434: Change tile provider on frontend submission form to current provider
Location question uses map tiles from carto.com; should be updated to use maptiler key via existing environment secret.

alexwolson:
Fixes #434.

- Uses the configured MapTiler API key for production raster tiles.
- Falls back to OpenStreetMap tiles when no MapTiler key is configured, so local development works without an account.
- Reuses the same tile configuration on the submission and dashboard maps, removing the CARTO and Thunderforest providers.
- Adds regression coverage for configured, missing, and empty API keys.

Tests:
- `npm test -- --runInBand` (53 passed)
- `npm run lint`
- `npm run build`

CI note: the API test job on this fork-based PR receives empty repository secrets and fails during database seeding because `SECURITY_PASSWORD_SALT` is unset; the frontend change does not touch the API.

### DIFF
--- a/bikespace_frontend/src/components/dashboard/map/Map.tsx
+++ b/bikespace_frontend/src/components/dashboard/map/Map.tsx
@@ -7,7 +7,7 @@ import {useStore} from '@/states/store';
 import {SidebarTab, useSidebarTab} from '@/states/url-params';
 import {useIsMobile} from '@/hooks/use-is-mobile';
 
-import {defaultMapCenter} from '@/utils/map-utils';
+import {defaultMapCenter, getRasterTileLayerProps} from '@/utils/map-utils';
 
 import {Spinner} from '@/components/shared-ui/spinner';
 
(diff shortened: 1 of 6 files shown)


tallcoleman (approved):
Thanks Alex!

Two optional additions if you're interested in taking a look:

- On our newer pages that use Maplibre with vector tiles, we have a backup tileset so that contributors get a local dev app that works without signing up for a Maptiler account (e.g. on [`ParkingMapPage`](https://github.com/bikespace/bikespace/blob/6166fd0bd149348ad428413bf8729e6e3de24cf1/bikespace_frontend/src/components/parking-map/parking-map-page/ParkingMapPage.tsx#L53)). For this case, maybe [tile.openstreetmap.org](https://operations.osmfoundation.org/policies/tiles/) would work?
- The `/dashboard` page uses raster tiles from another provider; they could be swapped to this one as well so that there's only one prod tile provider.

tallcoleman (approved):
Perfect!

You should be able to squash merge I think if you want to push it to prod. If you want to work on other issues in the future, just let me know and I can add you to the dev permissions group which should ensure that the ci tests all run correctly etc.

alexwolson:
Great! I don't have write permissions so I can't merge. I wanted to contribute this because I ran into this exact issue when submitting this morning - I don't think I'll be a regular contributor but figured I could do this one!

PR #437, opened 2026-09-12
Add docker compose API deployment and backup

### LINKED ISSUE #430: Improve deployment and backup options for the API
Description to come

Relates to https://github.com/bikespace/bikespace/issues/397

tallcoleman:
Adds configuration and documentation to allow the BikeSpace API to be deployed using Coolify (or regular docker compose) with automatic and configurable backups to S3-compatible storage. Closes #430 

Actually changing the prod deployment from fly.io to an API instance will be another PR with merge timing planned so that the new instance can have its database restored from prod just before the switch.

Key changes:

- Designed for deployment using Coolify but should also be compatible with vanilla docker compose
- Updated postgres version to 18 and pinned
- Added a backup system based on the one used for the [panoramax-coolify](https://github.com/bikespace/panoramax-coolify/tree/main/docker/full-keycloak-auth/backup) repo
- Added documentation

Tasks to do:

- [x] Update Postgres version
- [x] Add backup system
- [x] Confirm backup runs overnight
- [x] Test restore procedure
- [x] Test restore from prod
- [x] Check if updates need to be made to the ci (no, deploy ci action will be updated in follow up PR)
- [x] Check what cleanup should be made (swapped in this yaml instead of the old prodtest)
- [x] Check for documentation etc. that needs to be updated (in follow up PR: AGENTS.md, do a search for 'fly' to catch others)


sonalranjit (approved):
Looks awesome, there are a few coolify specific process that I am not familiar with but based on the docker compose commands and such it seems okay. These kind of changes we can't fully be sure until we actually perform it.

